PR #6, opened 2020-07-18
Force Python version to use

tetienne:
Several Python versions exist. We must agree on the minimum version we support. I propose the 3.6.8.


README.md
```diff
@@ -33,7 +33,7 @@ pip install git+https://github.com/iMicknl/python-tahoma-api.git@master#tahoma_a
 
 - Clone this repository
 - `cd python-tahoma-api`
-- Install the required Python version: `pyenv install 3.8.3`
+- Install the required Python version: `pyenv install`
```
tetienne, line 36: `pyenv install` use the version set in .python_version.

iMicknl: Is the pre-commit hook not requiring Python 3.8? That is what I remember from my struggles.

tetienne: Nope, it can run using the 3.6 also.

PR #8, opened 2020-07-18
Add .env support

tetienne:
When EMAIL and PASSWORD are found within a .env file, the associated values will bypass the parameters of the constructor.

### DIFF
--- a/tahoma_api/client.py
+++ b/tahoma_api/client.py
@@ -1,9 +1,11 @@
 """ Python wrapper for the Tahoma API """
+import os
 import urllib.parse
 from typing import Any, List, Optional
 
 import aiohttp
 import humps
+from dotenv import load_dotenv
 
 from tahoma_api.exceptions import BadCredentialsException, TooManyRequestsException
 from tahoma_api.models import Command, CommandMode, Device, State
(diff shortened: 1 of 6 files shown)


tahoma_api/client.py
```diff
@@ -14,16 +16,26 @@
 class TahomaClient:
     """ Interface class for the Tahoma API """
 
-    def __init__(self, username: str, password: str, api_url: str = API_URL) -> None:
+    def __init__(
+        self,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        api_url: str = API_URL,
```
tetienne, line 23: Is it really something important? Do you know other endpoints?

vlebourl: There's an issue opened in our component for that. https://github.com/iMicknl/ha-tahoma/issues/85

tahoma_api/client.py
```diff
@@ -14,16 +16,26 @@
...
+
+        Credentials can be set in .env file. Email and Password will be read from the
+        EMAIL and PASSWORD environment variables.
         """
 
-        self.username = username
-        self.password = password
+        load_dotenv()
+
+        self.username = os.getenv("EMAIL") or username
```
iMicknl, line 37: Is this something that should be in client? Imo this is something that belongs in a test script and not in the package itself. Especially with a generic environment variable name like 'email' and 'password', which can not be overriden. This could have a lot of side effects / issues.

tetienne: I made this because I saw your commend in your test file (which by the way, must be removed from this repository or put in an examples folder). 

You are right about the name of the variables. I can change them to TAHOMALINK_EMAIL and TAHOMALINK_PASSWORD. 

I can also make the hardcoded email and password priorities over the env.

tetienne: We can use a gist to share a test example, it will avoid polluting this repository. Feel free to reject this PR if you want.

iMicknl: If you agree, I would prefer moving the sample to an example folder and making sure this one won't be included in PyPi releases. Thus keeping the packaged code cleaner.

tetienne: Sure. I close this PR.

PR #9, opened 2020-07-18
Add roller shutter model

tetienne:


tetienne:
@iMicknl I will need help to send command. I still don't know how send them using the client.

tetienne:
The easiest would be to create a class for each qualified name or widget name instead of trying to centralize all the roller shutters in the same class. It will reduce a lot the complexity, and allow us to create a factory.

PR #13, opened 2020-07-20
Add scenarios support 

### LINKED ISSUE #11: Add scenario support
- [ ] Retrieve the scenarios: https://www.tahomalink.com/enduser-mobile-web/enduserAPI/actionGroups
- [ ] Run scenarios:  https://www.tahomalink.com/enduser-mobile-web/enduserAPI/exec/oid

tetienne:
Fix #11


tahoma_api/client.py
```diff
@@ -141,6 +141,16 @@ async def get_current_executions(self) -> List[Any]:
 
         return response
 
+    async def get_scenarios(self) -> List[Scenario]:
+        """ List the scenarios """
+        response = await self.__get("actionGroups")
+        return [Scenario(**scenario) for scenario in response]
+
+    async def launch_scenario(self, oid: str) -> str:
```
iMicknl, line 149: Can we change launch? I would like to align it to command as well in the end.

For example `execute_command` and `execute_scenario`?

tahoma_api/client.py
```diff
@@ -141,6 +141,16 @@ async def get_current_executions(self) -> List[Any]:
 
         return response
 
+    async def get_scenarios(self) -> List[Scenario]:
+        """ List the scenarios """
+        response = await self.__get("actionGroups")
```
iMicknl, line 146: I can understand that we don't need humps here, since it would slow down the request. However this will cause us having camelCase and snake_case throughout this file. Do we prefer millisecond performance hit over readability? Would be good to think about this for future PR's.

tetienne: For the moment, we only care about label and oid. So there is no need to transform the response :)

tahoma_api/client.py
tetienne: Sure, I will change this.

tetienne: Done.

tahoma_api/client.py
iMicknl: @tetienne nevermind, my reference was to `actionGroups`, however I missed that this is the GET endpoint...

PR #16, opened 2020-07-24
Add a context manager to TahomaClient

tetienne:
Will avoid to think about close session.


iMicknl:
Looks good! This would simplify the config flow for sure! 👍

PR #17, opened 2020-07-24
Create a States class to ease state access

tetienne:
Allow this kind of syntax:

```python
        devices = await client.get_devices()

        for device in devices:
            if device.states:
                if "core:RSSILevelState" in device.states:
                    print(device.states["core:RSSILevelState"].name)
                    print(device.states["core:RSSILevelState"].value)

```

### DIFF
--- a/tahoma_api/models.py
+++ b/tahoma_api/models.py
@@ -123,9 +123,9 @@ class CommandDefinitions:
         return self.__getitem__(name) is not None
 
     def __getitem__(self, command: str) -> Optional[CommandDefinition]:
-        return next(
-            (state for state in self._commands if state.command_name == command), None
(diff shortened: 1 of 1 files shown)


tetienne:
@iMicknl Code update, you should like :)

iMicknl:
> @iMicknl Code updated, you should like :)

LGTM! I see you marked it as a draft, so I will let you merge it yourself when you are ready :).

tetienne:
Yes I saw a potential issue for the command in operator. I have to check.

iMicknl:
@tetienne we should pin the version in the component and bump it here. Now an update to master can break the component.

Looks like it did now?

PR #18, opened 2020-07-27
Overwrite __len__ for CommandDefinitions and States

tetienne:


tetienne:
`asyncio.run` (only required for test.py) and  `from __future__ import annotations` are only available within 3.7. asyncio was introduced in 3.6, and improved in 3.7, so it does not seem to be an issue to change the minimal version we support.

PR #19, opened 2020-07-27
Clean up for first release

iMicknl:


README.md
```diff
@@ -14,7 +14,44 @@ This package is written for the Home Assistant [ha-tahoma](https://github.com/iM
...
+USERNAME = ""
+PASSWORD = ""
+
+async def main() -> None:
+    async with TahomaClient(USERNAME, PASSWORD) as client:
+        try:
+            await client.login()
+        except Exception as exception:  # pylint: disable=broad-except
+            print(exception)
+            return await client.close()
```
tetienne, line 37: You can  just do `return`, the `with` block will close the session automatically.

tetienne:
Once the 0.2 published, and the ha-tahoma integration updated to target it, I suggest to rename the package name from tahoma_api to just tahoma. See https://www.python.org/dev/peps/pep-0008/#package-and-module-names 

So 

```python
from tahoma_api.client import TahomaClient
from tahoma_api.exceptions import BadCredentialsException, TooManyRequestsException
``` 

would become

```python
from tahoma.client import TahomaClient
from tahoma.exceptions import BadCredentialsException, TooManyRequestsException
```

tetienne:
For the name of the project on pypi, _tahoma-api_ must be enough, we are on pypi, for sure it's a python project.

README.md
```diff
@@ -14,7 +14,44 @@ This package is written for the Home Assistant [ha-tahoma](https://github.com/iM
 ## Installation
 
 ```bash
-pip install git+https://github.com/iMicknl/python-tahoma-api.git@master#tahoma_api
+pip install python-tahoma-api
```
tetienne, line 17: `pip install tahoma-api

pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.poetry]
 name = "python-tahoma-api"
```
tetienne, line 2: `tahoma-api

README.md
iMicknl: @tetienne `tahoma-api` is already taken :(

tetienne: Ah yes indeed, what about *pyhoma*, as suggested by @vlebourl ?

pyproject.toml
```diff
@@ -1,20 +1,20 @@
...
+description = "Async Python wrapper to interact with internal Somfy TaHoma API"
 authors = ["Mick Vleeshouwer", "Vincent Le Bourlot", "Thibaut Etienne"]
 license = "MIT"
 readme = "README.md"
 homepage = "https://github.com/iMicknl/python-tahoma-api"
 repository = "https://github.com/iMicknl/python-tahoma-api"
 packages = [
-    { include = "tahoma_api" },
-    { include = "tahoma_api/**/*.py" },
+    { include = "tahoma" },
```
tetienne, line 11: We need this also? The next line is not enough?

tetienne: By the way, should we rename this package to pyhoma now or for the next release?

iMicknl: @tetienne I guess even the top one would be enough. I copied it from here: https://python-poetry.org/docs/pyproject/#packages. But I missed that they used `my_package` and `other_package`.

iMicknl: @tetienne I would prefer to change everything directly, before the first publish. We just need to hotfix the other component, but wouldn't be a big hassle.

tetienne: I think you can even remove the packages fields. Look at https://github.com/tetienne/somfy-open-api/blob/master/pyproject.toml. Don't use it and my package is ok: https://pypi.org/project/pymfy/#files

tetienne: OK, I let do it.

README.md
```diff
@@ -14,7 +14,44 @@ This package is written for the Home Assistant [ha-tahoma](https://github.com/iM
...
+pip install pyhoma
+```
+
+## Getting started
+
+```python
+import asyncio
+import time
+
+from tahoma.client import TahomaClient
```
tetienne, line 26: Just change this import to pyhoma, and I'm good.

iMicknl: Forgot to push again..

PR #23, opened 2020-07-28
fix get_current_execution

vlebourl:


tahoma_api/client.py
```diff
@@ -121,9 +121,9 @@ async def fetch_event_listener(self, listener_id: str) -> List[Event]:
     async def get_current_execution(self, exec_id: str) -> List[Execution]:
         """ Get an action group execution currently running """
         response = await self.__get(f"exec/current/{exec_id}")
-        executions = [Execution(**e) for e in humps.decamelize(response)]
+        execution = Execution(**humps.decamelize(response))
 
-        return executions
+        return execution
```
tetienne, line 126: You have also to change the returned type.

PR #25, opened 2020-07-29
Add unregister endpoint and automatically unregister on close

iMicknl:


tahoma_api/client.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, username: str, password: str, api_url: str = API_URL) -> None
 
         self.devices: List[Device] = []
         self.__roles = None
+        self.event_listeners: List[str] = []
```
tetienne, line 37: We can register only once. So a string is enough.

tahoma_api/client.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, username: str, password: str, api_url: str = API_URL) -> None
 
         self.devices: List[Device] = []
         self.__roles = None
+        self.event_listener: Optional[str] = None
```
tetienne, line 37: I would suggest self.listener_id. event_listener suggest it's an event object.

tahoma_api/client.py
```diff
@@ -118,6 +123,14 @@ async def fetch_event_listener(self, listener_id: str) -> List[Event]:
 
         return events
 
+    async def unregister_event_listener(self, listener_id: str) -> None:
+        """
+        Unregister an event listener
```
tetienne, line 128: Missing final dot.

tahoma_api/client.py
```diff
@@ -107,7 +107,7 @@ async def register_event_listener(self) -> str:
         """
         response = await self.__post("events/register")
         listener_id = response.get("id")
-        self.event_listener = listener_id
+        self.event_listener_id = listener_id
```
tetienne, line 110: IMO the variable listener_id is useless. But it's ok.

iMicknl: @tetienne I agree. However linter was complaining about the fact that it could be str or None. Instead of str.

Is there a way to annotate that response.get("id") will return a string always

tetienne: `.get` will return None if the key can not be found. Try with response["id"].

iMicknl: @tetienne same. I will keep it like this for now.

```
tahoma_api/client.py:111: error: Incompatible return value type (got "Optional[str]", expected "str")
```

PR #26, opened 2020-07-30
Set Connection to keep-alive by default

tetienne:


pyhoma/client.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, username: str, password: str, api_url: str = API_URL) -> None
         self.__roles = None
         self.event_listener_id: Optional[str] = None
 
-        self.session = aiohttp.ClientSession()
+        self.session = aiohttp.ClientSession(headers={"Connection": "keep-alive"})
```
iMicknl, line 39: Not related, but would it make sense to also add a User Agent and make this configurable on instantiation?

thijsdejong:
From what I can see in the documentation of aiohttp this does nothing. keep-alive is on by default when using a session.

`A session contains a connection pool inside. Connection reusage and keep-alives (both are on by default) may speed up total performance.` https://docs.aiohttp.org/en/stable/client_quickstart.html#make-a-request

Possibly tweaking the limit of connections in the connection pool yields better results?

PR #29, opened 2020-08-03
Dynamic state addition

vlebourl:
add support for dynamically adding state to States.

add support for default value to None

### DIFF
--- a/pyhoma/models.py
+++ b/pyhoma/models.py
@@ -134,7 +134,7 @@ class CommandDefinitions:
 class State:
     __slots__ = "name", "value", "type"
 
-    def __init__(self, name: str, value: Optional[str], type: int, **_: Any):
+    def __init__(self, name: str, type: int, value: Optional[str] = None, **_: Any):
         self.name = name
         self.value = value
         self.type = DataType(type)
--- a/pyhoma/models.py
+++ b/pyhoma/models.py
@@ -154,9 +154,11 @@ class States:
         return next((state for state in self._states if state.name == name), None)
 
     def __setitem__(self, name: str, state: State) -> None:
-        if self.__contains__(state.name):
-            return
-        self._states.append(state)
+        found = self.__getitem__(name)
+        if found is None:
+            self._states.append(state)
+        else:
+            self._states[self._states.index(found)] = state
 
     def __len__(self) -> int:
         return len(self._states)
--- a/pyhoma/models.py
+++ b/pyhoma/models.py
@@ -156,11 +156,7 @@ class States:
     def __setitem__(self, name: str, state: State) -> None:
         if self.__contains__(state.name):
             return
-        self._states.append(
-            State(
-                **{s: getattr(state, s) for s in state.__slots__ if hasattr(state, s)}
-            )
-        )
+        self._states.append(state)
 
     def __len__(self) -> int:
         return len(self._states)
--- a/pyhoma/models.py
+++ b/pyhoma/models.py
@@ -269,6 +269,7 @@ class ProductType(Enum):
 
 
 class DataType(Enum):
+    NONE = 0
     INTEGER = 1
     FLOAT = 2
     STRING = 3
--- a/pyhoma/models.py
+++ b/pyhoma/models.py
@@ -134,16 +134,10 @@ class CommandDefinitions:
 class State:
     __slots__ = "name", "value", "type"
 
-    def __init__(self, name: str, type: int, value: str = None, **_: Any):
+    def __init__(self, name: str, value: Optional[str], type: int, **_: Any):
         self.name = name
         self.value = value
         self.type = DataType(type)
-
-    def __getitem__(self, key):
-        return getattr(self, key)
-
-    def __setitem__(self, key, item):
-        self.__setattr__(key, item)
 
 
 class States:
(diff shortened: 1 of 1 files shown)


pyhoma/models.py
```diff
@@ -134,11 +134,17 @@ def __len__(self) -> int:
 class State:
     __slots__ = "name", "value", "type"
 
-    def __init__(self, name: str, value: str, type: int, **_: Any):
+    def __init__(self, name: str, type: int, value: str = None, **_: Any):
         self.name = name
         self.value = value
         self.type = DataType(type)
 
+    def __getitem__(self, key):
```
tetienne, line 142: Why add this method? It's for mimic the list behavior.

pyhoma/models.py
```diff
@@ -134,11 +134,17 @@ def __len__(self) -> int:
 class State:
     __slots__ = "name", "value", "type"
 
-    def __init__(self, name: str, value: str, type: int, **_: Any):
+    def __init__(self, name: str, type: int, value: str = None, **_: Any):
```
tetienne, line 137: Value is an optional string.

pyhoma/models.py
```diff
@@ -153,6 +159,13 @@ def __contains__(self, name: str) -> bool:
     def __getitem__(self, name: str) -> Optional[State]:
         return next((state for state in self._states if state.name == name), None)
 
+    def __setitem__(self, name: str, state: State) -> None:
```
tetienne, line 162: I don't think this is useful. We don't add state dynamically.

pyhoma/models.py
vlebourl: cause I thought it was needed by the States.__setitem__

tetienne: This is not required. You can remove it.

pyhoma/models.py
vlebourl: with 
```py
class State:
    __slots__ = "name", "value", "type"

    def __init__(self, name: str, value: Optional[str], type: int, **_: Any):
        self.name = name
        self.value = value
        self.type = DataType(type)
```
I still get the error: `__init__() missing 1 required positional argument: 'value'`
Defaulting value to None did solve that problem

pyhoma/models.py
```diff
@@ -153,6 +153,15 @@ def __contains__(self, name: str) -> bool:
...
+        if self.__contains__(state.name):
+            return
+        self._states.append(
+            State(
+                **{s: getattr(state, s) for s in state.__slots__ if hasattr(state, s)}
```
tetienne, line 161: I dont get it, why recreate a new state object instead of appending the given one?

vlebourl: image

pyhoma/models.py
```diff
@@ -153,6 +153,11 @@ def __contains__(self, name: str) -> bool:
     def __getitem__(self, name: str) -> Optional[State]:
         return next((state for state in self._states if state.name == name), None)
 
+    def __setitem__(self, name: str, state: State) -> None:
```
iMicknl, line 156: What is the use-case for this setter? And it can not be used to overwrite an existing state?

vlebourl: it allows to dynamically add a state if it was not present at the entity's initialization which happens for the Smart Thermostat.

pyhoma/models.py
```diff
@@ -153,6 +153,11 @@ def __contains__(self, name: str) -> bool:
...
+        if self.__contains__(state.name):
```
tetienne, line 157: That's seems unnatural, a set method must set a value, even if this one already exist.

vlebourl: although this ensure that an existing state can't be overwritten

tetienne: But if I want to do it? This is not the classic list behavior. @iMicknl  What do you think?

iMicknl: I don't see the scenario where you don't want to overwrite an existing state. If it is a valid scenario, I would do it in a separate function. Otherwise it is confusing if we override default behaviour.

vlebourl: fine by me

vlebourl: I think though that an existing value should be popped before the append, otherwise calling `device.states[state.name]`  will always refer to the first one in the list, and not the one appended.

vlebourl: it could be done with
```python
    def __setitem__(self, name: str, state: State) -> None:
        if self.__contains__(state.name):
            i = self._states.index(self.__getitem__(name))
            self._states.pop(i)
        self._states.append(state)
```

tetienne: Why not simply get the matching state with __getitem__, if None append the state to the list, else update the state value.

vlebourl: this works:
```py
    def __setitem__(self, name: str, state: State) -> None:
        s = self.__getitem__(name)
        if s is None:
            self._states.append(state)
        else:
            i = self._states.index(self.__getitem__(name))
            self._states[i] = state
```

tetienne: Almost perfect. Just rewrite the I assignment to self._states.index(s)

vlebourl: indeed... again with the Shadocks 😋

PR #32, opened 2020-08-06
Make ui_class optional

tetienne:
Fix https://github.com/iMicknl/ha-tahoma/issues/205


tetienne:
@vlebourl That's an edge case, the device returned is marked as "not available". On HA side, we already handle this case. We map the None value... to nothing. We will see if "real" device will come with this behavior.

Issue #35, opened 2020-08-14
'Attempt to decode JSON with unexpected mimetype: '


thijsdejong:
I'm currently unable to open my covers using Home Assistant. I did not restart yet since I thought the errors might be useful to handle in the future.
```
0, message='Attempt to decode JSON with unexpected mimetype: ', url=URL('https://tahomalink.com/enduser-mobile-web/enduserAPI/exec/apply')
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 125, in handle_call_service
    await hass.services.async_call(
  File "/usr/src/homeassistant/homeassistant/core.py", line 1308, in async_call
    task.result()
  File "/usr/src/homeassistant/homeassistant/core.py", line 1343, in _execute_service
    await handler.func(service_call)
  File "/usr/src/homeassistant/homeassistant/helpers/entity_component.py", line 208, in handle_service
    await self.hass.helpers.service.entity_service_call(
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 454, in entity_service_call
    future.result()  # pop exception if have
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 583, in async_request_call
    await coro
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 485, in _handle_entity_call
    await result
  File "/config/custom_components/tahoma/cover.py", line 199, in async_open_cover
    await self.async_execute_command(
  File "/config/custom_components/tahoma/tahoma_device.py", line 155, in async_execute_command
    await self.coordinator.client.execute_command(
  File "/usr/local/lib/python3.8/site-packages/pyhoma/client.py", line 157, in execute_command
    return await self.execute_commands(device_url, [command], label)
  File "/usr/local/lib/python3.8/site-packages/pyhoma/client.py", line 170, in execute_commands
    response = await self.__post("exec/apply", payload)
  File "/usr/local/lib/python3.8/site-packages/pyhoma/client.py", line 196, in __post
    await self.check_response(response)
  File "/usr/local/lib/python3.8/site-packages/pyhoma/client.py", line 209, in check_response
    result = await response.json()
  File "/usr/local/lib/python3.8/site-packages/aiohttp/client_reqrep.py", line 1026, in json
    raise ContentTypeError(
[...]

PR #36, opened 2020-08-14
Fix case where error response does not return a JSON

tetienne:
See #35

### DIFF
--- a/pyhoma/client.py
+++ b/pyhoma/client.py
@@ -100,7 +100,7 @@ class TahomaClient:
         """
         Retrieve states of requested device
         """
-        response = await self.__get(
+        response = await self.__get_and_retry(
             f"setup/devices/{urllib.parse.quote_plus(deviceurl)}/states"
         )
         state = [State(**s) for s in humps.decamelize(response)]
@@ -145,14 +145,14 @@ class TahomaClient:
 
     async def get_current_execution(self, exec_id: str) -> Execution:
         """ Get an action group execution currently running """
-        response = await self.__get(f"exec/current/{exec_id}")
+        response = await self.__get_and_retry(f"exec/current/{exec_id}")
         execution = Execution(**humps.decamelize(response))
 
         return execution
 
     async def get_current_executions(self) -> List[Execution]:
         """ Get all action groups executions currently running """
-        response = await self.__get("exec/current")
+        response = await self.__get_and_retry("exec/current")
         executions = [Execution(**e) for e in humps.decamelize(response)]
 
         return executions
@@ -183,17 +183,17 @@ class TahomaClient:
             "label": label,
             "actions": [{"deviceURL": device_url, "commands": commands}],
         }
-        response = await self.__post("exec/apply", payload)
+        response = await self.__post_and_retry("exec/apply", payload)
         return response["execId"]
 
     async def get_scenarios(self) -> List[Scenario]:
         """ List the scenarios """
-        response = await self.__get("actionGroups")
+        response = await self.__get_and_retry("actionGroups")
         return [Scenario(**scenario) for scenario in response]
 
     async def execute_scenario(self, oid: str) -> str:
         """ Execute a scenario """
-        response = await self.__post(f"exec/{oid}")
+        response = await self.__post_and_retry(f"exec/{oid}")
         return response["execId"]
 
     async def __get(self, endpoint: str) -> Any:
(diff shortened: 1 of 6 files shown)


pyhoma/client.py
```diff
@@ -206,7 +206,12 @@ async def check_response(response: ClientResponse) -> None:
         #  'error': 'Too many requests, try again later : login with xxx@xxx.tld'}
         #  'error': 'Bad credentials'}
         #  'error': 'Your setup cannot be accessed through this application'}
-        result = await response.json()
+        try:
+            result = await response.json()
+        except ContentTypeError:
+            body = await response.text()
+            raise Exception(body or "Something bad happened")
```
iMicknl, line 213: For now raising an exception is fine, until we understand what the content is. Then I would like to handle it in a better way.

Can we change `Something bad happened` to a bit more descriptive error perhaps?

iMicknl (approved):

pyhoma/client.py
tetienne: What do you think about adding the URL, payload and status code?

vlebourl: Agreed with a more explicit message

vlebourl (approved):

tetienne:
@iMicknl @vlebourl I've ease again the error management logic. With the previous process, when a command on a wrong deviceurl was sent, the returned error was ignored. It's an error 400 with the following JSON: `{'errorCode': 'INVALID_FIELD_VALUE', 'error': 'Invalid device URL : foo'}

pyhoma/client.py
```diff
@@ -206,23 +268,24 @@ async def check_response(response: ClientResponse) -> None:
...
-                message = result.get("error")
-
-                if "Too many requests" in message:
-                    raise TooManyRequestsException(message)
+        try:
+            result = await response.json(content_type=None)
+        except JSONDecodeError:
+            body = await response.text()
+            raise Exception(
+                f"Something bad happened when requesting {response.url}. {body}"
```
iMicknl, line 276: Will this actually fix #35? Or was the fix to add `content_type=None`, since the TaHoma API sometimes adds the wrong mimetype?

pyhoma/client.py
```diff
@@ -206,23 +268,24 @@ async def check_response(response: ClientResponse) -> None:
...
+            raise TooManyRequestsException(message)
 
-                raise Exception(message)
+        if message == "Bad credentials":
+            raise BadCredentialsException(message)
 
-        if 400 < response.status < 500:
-            message = result.get("error")
+        # {"errorCode":"RESOURCE_ACCESS_DENIED","error":"Not authenticated"}
+        if message == "Not authenticated":
```
iMicknl, line 288: Good catch!

pyhoma/client.py
tetienne: The one fixing #35 is `content_type=None`. It avoid aiohttp to look for the mimetype. It's why now I have to check for JSONDecodeError in case JSON is not returned.

iMicknl (changes requested):
Could we split this PR in two? One PR that contains the JSON decoding fix for #35 and the NotAuthenticatedException + another PR adding all the Event Listener logic?

I am not sure yet if I like the behaviour where we hardcode all the event handling in the api library, instead of the clients leveraging this library. I rather add some helper functions in this library, available for clients to consume.

pyhoma/client.py
```diff
@@ -52,50 +65,83 @@ async def __aexit__(
...
+            if not event.deviceurl:
+                continue
+            device = devices[event.deviceurl]
+            device.states[state.name] = state  # type: ignore
+            device.states[state.name].value = self._cast_state(state)  # type: ignore
+
+    @staticmethod
+    def _update_available(devices: Dict[str, Device], event: Event) -> None:
+        if event.name == "DeviceAvailableEvent" and event.deviceurl:
+            devices[event.deviceurl].available = True
```
tetienne, line 124: @iMicknl You should recognize this code :)

tetienne:
@iMicknl No problem, I can remove the fetch_device method. Do you want me also remove the retry logic?
Having some logic within the library around the event, it's IMO something good. It will avoid the clients to recreate the wheel each time. But we can rework what I did in another PR as you suggest.

iMicknl:
> @iMicknl No problem, I can remove the fetch_device method. Do you want me also remove the retry logic?
> Having some logic within the library around the event, it's IMO something good. It will avoid the clients to recreate the wheel each time. But we can rework what I did in another PR as you suggest.

It would be great if you could split them in 3 for now.
1. fix JSON decode issue
2. retry logic
3. reconnect event listener etc.

iMicknl:
@tetienne can we close this one? I think we cherry picked all your changes in different PR's?

PR #37, opened 2020-08-22
Support adding a custom session

iMicknl:
```python3
self.websession = async_create_clientsession(
	self.hass, cookie_jar=aiohttp.CookieJar(unsafe=True)
)
```

In order to achieve platinum status, HA requires to pass a Client Session to the library. I made it optional, since adding your own session won't be necessary in many cases.

### DIFF
--- a/pyhoma/client.py
+++ b/pyhoma/client.py
@@ -20,12 +20,20 @@ API_URL = "https://tahomalink.com/enduser-mobile-web/enduserAPI/"  # /doc for AP
 class TahomaClient:
     """ Interface class for the Tahoma API """
 
-    def __init__(self, username: str, password: str, api_url: str = API_URL) -> None:
+    def __init__(
+        self,
+        username: str,
+        password: str,
+        api_url: str = API_URL,
+        session: aiohttp.ClientSession = None,
+    ) -> None:
         """
         Constructor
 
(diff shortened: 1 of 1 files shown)


pyhoma/client.py
```diff
@@ -20,12 +20,20 @@
 class TahomaClient:
     """ Interface class for the Tahoma API """
 
-    def __init__(self, username: str, password: str, api_url: str = API_URL) -> None:
+    def __init__(
+        self,
+        username: str,
+        password: str,
+        api_url: str = API_URL,
+        session: aiohttp.ClientSession = None,
```
tetienne, line 28: Can you just write `session: ClientSession = None,`? and import the right thing? It's not the style we use in the code currently.

tetienne:
Your modification makes sense. It can be useful even if uses outside HA.

pyhoma/client.py
iMicknl: Should I change it on line 39 as well? Before my modification, it was already written as `aiohttp.ClientSession()`

tetienne: If you can yes.

iMicknl: Fixed!

tetienne (approved):

PR #38, opened 2020-08-22
Add None to ProductType Enum

iMicknl:
To fix issues with faulty devices.

### DIFF
--- a/pyhoma/models.py
+++ b/pyhoma/models.py
@@ -258,6 +258,7 @@ class Scenario:
 
 
 class ProductType(Enum):
+    NONE = 0
     ACTUATOR = 1
     SENSOR = 2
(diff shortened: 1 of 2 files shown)


pyhoma/models.py
```diff
@@ -258,6 +258,7 @@ def __init__(self, label: str, oid: str, **_: Any):
 
 
 class ProductType(Enum):
+    NONE = 0
```
tetienne, line 261: Somfy can return 0 as type value?

iMicknl: https://github.com/iMicknl/ha-tahoma/issues/205#issuecomment-677771811 :(

tetienne (approved):

PR #39, opened 2020-08-22
Add support for cancelling executions

iMicknl:
Not sure if I should call it `cancel_command` or `cancel_execution`?

### DIFF
--- a/pyhoma/client.py
+++ b/pyhoma/client.py
@@ -200,11 +200,10 @@ class TahomaClient:
             await self.check_response(response)
             return await response.json()
 
-    async def __delete(self, endpoint: str) -> Any:
+    async def __delete(self, endpoint: str) -> None:
         """ Make a DELETE request to the TaHoma API """
         async with self.session.delete(f"{self.api_url}{endpoint}") as response:
             await self.check_response(response)
-            return await response.json()
 
     @staticmethod
     async def check_response(response: ClientResponse) -> None:
--- a/pyhoma/client.py
+++ b/pyhoma/client.py
@@ -156,6 +156,10 @@ class TahomaClient:
             command = Command(command)
         return await self.execute_commands(device_url, [command], label)
 
+    async def cancel_command(self, exec_id: str) -> None:
+        """ Cancel a running setup-level execution """
+        await self.__delete(f"/exec/current/setup/{exec_id}")
+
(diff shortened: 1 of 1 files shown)


tetienne (approved):

tetienne:
_cancel_command_ is fine IMO, as we _execute_command_

vlebourl:
> _cancel_command_ is fine IMO, as we _execute_command_
fine by me as well

pyhoma/client.py
```diff
@@ -196,6 +200,12 @@ async def __post(
             await self.check_response(response)
             return await response.json()
 
+    async def __delete(self, endpoint: str) -> Any:
+        """ Make a POST request to the TaHoma API """
```
vlebourl, line 204: Isn't that a DELETE request?

vlebourl (approved):

pyhoma/client.py
tetienne: Good catch :D

pyhoma/client.py
```diff
@@ -196,6 +200,12 @@ async def __post(
...
+        async with self.session.delete(f"{self.api_url}{endpoint}") as response:
+            await self.check_response(response)
+            return await response.json()
```
tetienne, line 207: The DELETE command does not return anything.

iMicknl: @tetienne but it can return a status code in the range of 400/500? That's why I kept the `check_response` check.

iMicknl: Sorry, I missed that you pointed to this line `return await response.json()`... DELETE responses can return a JSON response, however in this case they don't. Should I remove it or keep it for other (potential) use-cases.

tetienne: I suggest you to call raise_for_status: https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse.raise_for_status and return nothing.

iMicknl: @tetienne but in that case we won't be able to throw exceptions like 'Unauthorized' etc? I would prefer to keep `await self.check_response(response)`, but I can remove `return await response.json()` for now?

tetienne: Indeed, remove `             return await response.json() ` will be enough.

PR #43, opened 2020-08-27
Relogin when disconnected

tetienne:


README.md
```diff
@@ -46,7 +46,7 @@ async def main() -> None:
         listener_id = await client.register_event_listener()
 
         while True:
-            events = await client.fetch_event_listener(listener_id)
+            events = await client.fetch_events()
```
tetienne, line 49: We fetch "events" not a "listener".

pyhoma/client.py
```diff
@@ -21,6 +22,10 @@
 API_URL = "https://tahomalink.com/enduser-mobile-web/enduserAPI/"  # /doc for API doc
 
 
+async def relogin(invocation: Dict[str, Any]) -> None:
+    await invocation["args"][0].login()
```
iMicknl, line 26: What is the difference between this and just calling `login`?

README.md
iMicknl: True, but this is a breaking change. I don't think that anyone is using this package yet, but would be good to mention this somewhere in the changelog.

pyhoma/client.py
tetienne: backliff needs a callback method with a specific signature.

pyhoma/client.py
```diff
@@ -123,40 +130,49 @@ async def register_event_listener(self) -> str:
...
-    async def fetch_event_listener(self, listener_id: str) -> List[Event]:
+    async def fetch_events(self) -> List[Event]:
         """
         Fetch new events from a registered event listener. Fetched events are removed
         from the listener buffer. Return an empty response if no event is available.
         Per-session rate-limit : 1 calls per 1 SECONDS period for this particular
         operation (polling)
         """
-        response = await self.__post(f"events/{listener_id}/fetch")
+        response = await self.__post(f"events/{self.event_listener_id}/fetch")
```
iMicknl, line 140: What if `fetch_events` does throw a `NotAuthenticatedException`? How do we register the event listener for future executions from `ha-tahoma`?

pyhoma/client.py
iMicknl: Looks ugly but backoff library looks like a good solution 👍🏻

pyhoma/client.py
tetienne: Yes it what you ask for. It's why in #36 I created a `fetch_devices` method.

iMicknl: Looks good, just wanted to understand the flow.

pyhoma/client.py
```diff
@@ -77,10 +82,9 @@ async def login(self) -> bool:
         response = await self.__post("login", data=payload)
 
         if response.get("success"):
-            self.__roles = response.get("roles")
-
+            if register:
```
iMicknl, line 85: Can we change `register` to something which is more descriptive? `register_event_listener` or something similar?

tetienne: I hesitate, I don't like long name. But you are right.

pyhoma/client.py
```diff
@@ -123,40 +130,49 @@ async def register_event_listener(self) -> str:
...
         operation (polling)
         """
-        response = await self.__post(f"events/{listener_id}/fetch")
+        response = await self.__post(f"events/{self.event_listener_id}/fetch")
         events = [Event(**e) for e in humps.decamelize(response)]
 
         return events
 
-    async def unregister_event_listener(self, listener_id: str) -> None:
+    async def unregister_event_listener(self) -> None:
```
iMicknl, line 145: Should `unregister_event_listener` and `register_event_listener` not use backoff?

pyhoma/client.py
```diff
@@ -96,6 +100,9 @@ async def get_devices(self, refresh: bool = False) -> List[Device]:
 
         return devices
 
+    @backoff.on_exception(
+        backoff.expo, NotAuthenticatedException, max_tries=2, on_backoff=relogin
```
iMicknl, line 104: In general, should we not reconnect on `DisconnectedServerError` as well?

pyhoma/client.py
tetienne: If you are logout, you don't need to unregister the listener as this one has been deleted. 
About register, normally you do it just after a login, it's why I didn't add a retry logic.

pyhoma/client.py
tetienne: I'm not sure we can get this error now we handle properly the logout.

pyhoma/client.py
iMicknl: Makes sense, let's see how it works like this. We can always add it if we face issues.

iMicknl (approved):

PR #44, opened 2020-08-29
Add refresh states command

iMicknl:
Needed for https://github.com/iMicknl/ha-tahoma/issues/167

### DIFF
--- a/pyhoma/client.py
+++ b/pyhoma/client.py
@@ -114,6 +114,15 @@ class TahomaClient:
 
         return state
 
+    @backoff.on_exception(
+        backoff.expo, NotAuthenticatedException, max_tries=2, on_backoff=relogin
+    )
+    async def refresh_states(self) -> None:
+        """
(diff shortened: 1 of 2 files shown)


pyhoma/client.py
```diff
@@ -114,6 +114,15 @@ async def get_state(self, deviceurl: str) -> List[State]:
...
 
+    @backoff.on_exception(
+        backoff.expo, NotAuthenticatedException, max_tries=2, on_backoff=relogin
+    )
+    async def refresh_states(self) -> None:
+        """
+        Ask the box to refresh all devices states for protocols supporting that operation
+        """
+        await self.__post("/setup/devices/states/refresh")
+
```
tetienne, line 125: You don't want to check the response?

iMicknl: Response is `{}` and `self.__post` checks the response for errors already.

PR #47, opened 2020-09-17
Add gateway endpoint and Enums for Event class

iMicknl:


pyhoma/models.py
```diff
@@ -277,3 +389,245 @@ class DataType(Enum):
...
+    UNKNOWN_DETAILED_ERROR = 10007
+
+
+class UpdateBoxStatus(Enum):
+    NOT_UPDATABLE = "NOT_UPDATABLE"
+    READY_TO_UPDATE = "READY_TO_UPDATE"
+    READY_TO_BE_UPDATED_BY_SERVER = "READY_TO_BE_UPDATED_BY_SERVER"
+    READY_TO_UPDATE_LOCALLY = "READY_TO_UPDATE_LOCALLY"
+    UP_TO_DATE = "UP_TO_DATE"
+    UNKNOWN = "UNKNOWN"
```
tetienne, line 633: It's nice to have these enums. But where should we use them?

iMicknl: These enums are used by `Gateway` and could be used to check if there is an update available. Some of the other enums are not for Gateway and need to be mapped in future PR's. I just wanted to include them already before I lose the source again :D.

iMicknl: @tetienne included a few already. Would make our coordinator code a bit more strongly typed using those enums.

tetienne: That's great!

pyhoma/models.py
```diff
@@ -197,30 +198,81 @@ class Event:
...
+        deleted_raw_devices_count: Optional[Any] = None,
+        protocol_type: Optional[Any] = None,
         gateway_id: Optional[str] = None,
         exec_id: Optional[str] = None,
         deviceurl: Optional[str] = None,
         device_states: Optional[List[Dict[str, Any]]] = None,
-        old_state: Optional[str] = None,
-        new_state: Optional[str] = None,
+        old_state: Optional[ExecutionState] = None,
+        new_state: Optional[ExecutionState] = None,
```
tetienne, line 242: So many cases possible :o

pyhoma/models.py
tetienne: This class is now really huge. What do you think about moving all the enums to a dedicated module?

iMicknl: > This class is now really huge. What do you think about moving all the enums to a dedicated module?

Should I create an `enums.py`?

tetienne (approved):

PR #48, opened 2020-10-05
Improve retry on error logic

iMicknl:


pyhoma/client.py
```diff
@@ -193,6 +205,7 @@ async def get_current_executions(self) -> List[Execution]:
 
         return executions
 
+    @backoff.on_exception(backoff.expo, TooManyExecutionsException, max_tries=10)
```
tetienne, line 208: Can you check if we cannot add a timeout? Because I'm afraid that if we retried 10 times too quickly we will still the too many requests errors.

iMicknl: I was first looking into tweaking this, but `backoff.expo` does this already. The 10 times are scattered over 2 minutes, thus should be fine.

tetienne: Perfect!

tetienne (approved):

PR #50, opened 2020-10-18
Fix for unknown failure_type_code

### LINKED ISSUE #49: Error fetching TaHoma Event Fetcher data: 8301 is not a valid FailureType
```
Error fetching TaHoma Event Fetcher data: 8301 is not a valid FailureType
```

We should add more FailureTypes, or at least make sure we try/catch them.

iMicknl:
Fixes #49

### DIFF
--- a/pyhoma/models.py
+++ b/pyhoma/models.py
@@ -289,8 +289,6 @@ class Event:
             )
         except ValueError:
             self.failure_type_code = failure_type_code
-            print(failure_type_code)
-            print(failure_type)
 
 
 class Execution:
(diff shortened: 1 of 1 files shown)


pyhoma/models.py
```diff
@@ -285,6 +283,15 @@ def __init__(
...
             self.name = name
 
+        try:
+            self.failure_type_code = (
+                FailureType(failure_type_code) if failure_type_code else None
+            )
+        except ValueError:
+            self.failure_type_code = failure_type_code
+            print(failure_type_code)
+            print(failure_type)
```
tetienne, line 293: Instead of a print, we should use a logger.

iMicknl: Oops, this was my debug code...

PR #51, opened 2020-10-23
Add get_device_definition to retrieve the dict of all device definition.

vlebourl:
This will allow adding a service to ha-tahoma to retrieve definitions such
as available commands, useful to discover commands to be run by the execute_command
service.

### DIFF
--- a/pyhoma/client.py
+++ b/pyhoma/client.py
@@ -137,7 +137,7 @@ class TahomaClient:
             f"setup/devices/{urllib.parse.quote_plus(deviceurl)}"
         )
 
-        return response['definition']
+        return response["definition"]
 
     @backoff.on_exception(
         backoff.expo, NotAuthenticatedException, max_tries=2, on_backoff=relogin
--- a/pyhoma/client.py
+++ b/pyhoma/client.py
@@ -129,7 +129,7 @@ class TahomaClient:
     @backoff.on_exception(
         backoff.expo, NotAuthenticatedException, max_tries=2, on_backoff=relogin
     )
-    async def get_device_definition(self, deviceurl: str) -> Dict:
+    async def get_device_definition(self, deviceurl: str) -> Dict[str, Any]:
         """
         Retrieve a particular setup device definition
         """
--- a/pyhoma/client.py
+++ b/pyhoma/client.py
@@ -129,7 +129,7 @@ class TahomaClient:
     @backoff.on_exception(
         backoff.expo, NotAuthenticatedException, max_tries=2, on_backoff=relogin
     )
-    async def get_device_definition(self, deviceurl: str) -> Dict[str, Any]:
+    async def get_device_definition(self, deviceurl: str) -> JSON:
         """
         Retrieve a particular setup device definition
         """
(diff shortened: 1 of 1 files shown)


vlebourl:
This could be restricted to return the lis of commands instead of the full `definition` dict.

tetienne (approved):

tetienne:
Nice addition!

pyhoma/client.py
```diff
@@ -126,6 +126,19 @@ async def get_gateways(self, refresh: bool = False) -> List[Gateway]:
...
+    )
+    async def get_device_definition(self, deviceurl: str) -> JSON:
+        """
+        Retrieve a particular setup device definition
+        """
+        response = await self.__get(
+            f"setup/devices/{urllib.parse.quote_plus(deviceurl)}"
+        )
+
+        return response["definition"]
```
iMicknl, line 140: Is definition always present? And what if we try to pull the definition for an unknown device or Hue/Zwave device? Are those scenarios tested and handled.

pyhoma/client.py
```diff
@@ -126,6 +126,19 @@ async def get_gateways(self, refresh: bool = False) -> List[Gateway]:
 
         return gateways
 
+    @backoff.on_exception(
+        backoff.expo, NotAuthenticatedException, max_tries=2, on_backoff=relogin
+    )
+    async def get_device_definition(self, deviceurl: str) -> JSON:
```
iMicknl, line 132: Is `JSON` a valid return type? We use Dict for all others, but I am not sure why. Do you know @tetienne ?

iMicknl (commented):
Nice addition!

pyhoma/client.py
vlebourl: - don't know
- don't know
- no

pyhoma/client.py
vlebourl: JSON is defined as 
```py
JSON = Union[Dict[str, Any], List[Dict[str, Any]]]
```

vlebourl: It just makes it clearer IMO

pyhoma/client.py
vlebourl: I can change to `response.get("definition")`, that should keep us safe from "definition" not being present.

vlebourl: fixed in https://github.com/iMicknl/python-tahoma-api/pull/51/commits/a5f12e5b4db6c0863d11c164778024f1071d2386

pyhoma/client.py
tetienne: The guy who wrote that must be really nice :D

pyhoma/client.py
```diff
@@ -126,6 +126,19 @@ async def get_gateways(self, refresh: bool = False) -> List[Gateway]:
...
+        """
+        Retrieve a particular setup device definition
+        """
+        response = await self.__get(
+            f"setup/devices/{urllib.parse.quote_plus(deviceurl)}"
+        )
+
+        return response.get("definition")
```
tetienne, line 140: So finally, the return  type is `Optional[JSON]`

pyhoma/client.py
iMicknl: Maybe we should have a look at other definitions as well! Good catch.

iMicknl (approved):

PR #53, opened 2020-10-24
Handle maintenance error

### LINKED ISSUE #52: Add `ServerIsDownForMaintenance` exception, catch 503 statuscode
Apparently they update their service quite often lately, with some planned downtime. Would be good to catch this error gracefully, instead of spamming the log with their HTML page.

```
2020-10-23 06:11:34 DEBUG (MainThread) [custom_components.tahoma.coordinator] Unknown error while requesting https://tahomalink.com/enduser-mobile-web/enduserAPI/events/4ffb38c0-ac10-3401-4cd1-e427614e1f4c/fetch. 503 - <html>
<head>
  <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
  <title>Overkiz - Universal HomeControl</title>
  <!-- PROD1 -->
  <style>
   body {
        font-family: Arial, Verdana, Helvetica, sans-serif;
	font-size: 10pt;
	color: black;
   } 
  </style>
</head>
<body>
<div align="center">
<br/><br/><br/><br/>
<table border="0" width="90%">
  <tr>
    <td width="50%" align="center"><img src="/error/overkiz_logo.gif"/></td>
    <td>
      <p class="intro">
        <b>Server is down for maintenance</b><br/>
        <br/>
        We are sorry for the inconvenience.<br/>
        Please try again in a few minutes.
      </p>
    </td>
  </tr>
</table>
</div>
</body>

</html>
```

tetienne:
Fix #52

### DIFF
--- a/pyhoma/client.py
+++ b/pyhoma/client.py
@@ -295,7 +295,7 @@ class TahomaClient:
             result = await response.json(content_type=None)
         except JSONDecodeError:
             result = await response.text()
-            if "maintenance" in result:
+            if "Server is down for maintenance" in result:
                 raise MaintenanceException("Server is down for maintenance")
             raise Exception(
                 f"Unknown error while requesting {response.url}. {response.status} - {result}"
--- a/pyhoma/client.py
+++ b/pyhoma/client.py
@@ -12,6 +12,7 @@ from aiohttp import ClientResponse, ClientSession, ServerDisconnectedError
 
 from pyhoma.exceptions import (
     BadCredentialsException,
+    MaintenanceException,
     NotAuthenticatedException,
     TooManyExecutionsException,
     TooManyRequestsException,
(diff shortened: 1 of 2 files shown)


pyhoma/client.py
```diff
@@ -294,6 +295,8 @@ async def check_response(response: ClientResponse) -> None:
             result = await response.json(content_type=None)
         except JSONDecodeError:
             result = await response.text()
+            if "maintenance" in result:
```
iMicknl, line 298: My preference would be to match on `Server is down for maintenance`? Just to be sure we don't catch an edge case where `maintenance` is somewhere in the report.

pyhoma/client.py
```diff
@@ -294,6 +295,8 @@ async def check_response(response: ClientResponse) -> None:
...
+                raise MaintenanceException("Server is down for maintenance")
```
iMicknl, line 299: Could we possibly also check for 503 status code? At least Somfy is using the correct status code this time. `503 Service Unavailable.`

tetienne: I don't think we have to check the status code again. We are already in the error case.

iMicknl: I am fine with the current implementation. We could also do it different, like this:

catch 503 error
 if 503 and "maintenance" -> throw maintenance
 otherwise throw ServiceUnavailableException

Hopefully we don't need the last one, but otherwise we throw a good exception at least.

iMicknl (approved):

PR #56, opened 2020-12-07
Enable dependabot

tetienne:


iMicknl (approved):
Nice! Do we need to have the integration enabled or will this enable it directly?

tetienne:
@iMicknl it will be done automatically.

PR #68, opened 2021-01-12
Fix connectivity model

iMicknl:


iMicknl:
Really small fix, self-approving.

PR #75, opened 2021-01-28
Add endpoint to list Execution History

iMicknl:
```
        history = await client.get_execution_history()

        for item in history:
            timestamp = datetime.datetime.fromtimestamp(int(item.event_time) / 1000)

            for command in item.commands:
                date = timestamp.strftime("%Y-%m-%d %H:%M:%S")

                print(
                    "{timestamp}: {command} executed via {app} on {device}, with {parameters}.".format(
                        command=command.command,
                        timestamp=date,
                        device=command.deviceurl,
                        parameters=command.parameters,
                        app=item.label,
                    )
                )
```

### DIFF
--- a/pyhoma/client.py
+++ b/pyhoma/client.py
@@ -18,7 +18,16 @@ from pyhoma.exceptions import (
     TooManyExecutionsException,
     TooManyRequestsException,
 )
-from pyhoma.models import Command, Device, Event, Execution, Gateway, Scenario, State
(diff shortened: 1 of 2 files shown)


iMicknl:
@vlebourl this can be used so that users don't need to struggle with Chrome Developer Panel. And it can be used by non TaHoma users.

vlebourl (approved):

PR #93, opened 2021-02-22
Add support for places endpoint

iMicknl:
- Add places oid to device 
- Adds places endpoint


iMicknl (commented):
Self-approving for now, please feel free to add comments. I will address them in a  future PR.

PR #134, opened 2021-05-27
Add Hi Kumo WiFi Adapter as GatewayType

iMicknl:


pyhoma/enums.py
```diff
@@ -17,6 +17,7 @@ class GatewayType(Enum):
     TAHOMA = 15
     VERISURE_ALARM_SYSTEM = 20
     KIZBOX_MINI = 21
+    HI_KUMO_ADAPTER = 22  # Hi Kumo Adapter SPX-WFG01 (constant added manually)
```
tetienne, line 20: Why this comment?

iMicknl: All others come from a list with constants from tahomalink.com, thus use their enums. The other I added for Hi Kumo has a similar comment, so we know later on that this was added by us.

PR #154, opened 2021-07-21
Default to empty list instead of None

tetienne:
We often iterate over states or attributes, to ease client code, I suggest to return an empty list instead of None when the data does not exist. It will avoid None check.

### DIFF
--- a/pyhoma/models.py
+++ b/pyhoma/models.py
@@ -61,14 +61,14 @@ class Device:
         **_: Any
     ) -> None:
         self.id = deviceurl
-        self.attributes = States(attributes) if attributes else None
+        self.attributes = States(attributes) if attributes else []
         self.available = available
         self.definition = Definition(**definition)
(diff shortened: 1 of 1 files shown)


iMicknl:
@tetienne do we need to list this as a breaking change? Makes sense to simplify this and always return the same type.

tetienne:
I don’t see cases where it would break stuff. Most of the time you want a specific state so you test your list. But mypy will still complain in some cases as we still declare the variables as optional and setting an empty array as default parameter is not a good practice :S

iMicknl:
> But mypy will still complain in some cases as we still declare the variables as optional and setting an empty array as default parameter is not a good practice :S

Hmm, does this mean that we can't merge this one yet? Or is this PR ready to merge :-)

tetienne:
Well that's OK for now.

iMicknl (approved):

PR #155, opened 2021-07-22
Add a list of supported Overkiz servers as constant

iMicknl:


tetienne (commented):
You cannot use an enum ?

iMicknl:
> You cannot use an enum ?

That was quick! I don't think you can use an ENUM with multiple values? And I didn't like the approach of ha-tahoma a lot, where we have multiple lists that share the same key. (endpoint + manufacturer)

tetienne:
Ah yes indeed. Sorry for enum I still think about Java where they are very powerful.

tetienne (approved):

iMicknl:
> Ah yes indeed. Sorry for enum I still think about Java where they are very powerful.

Makes sense, in Python you would need an external package to simulate the same.

However, I will just try to see if I can get it working in HA config flow before I merge this one. It seems that it is not possible to work with key/value pairs in the config flow dropdown.

```python
        vol.Required(CONF_HUB, default=self._default_hub): vol.In(
                        SUPPORTED_HUBS.keys()
                    ),
```

iMicknl:
It makes more sense to have them as NamedTuple, instead of dataclass. See https://github.com/home-assistant/core/pull/53341/files.

PR #184, opened 2021-09-23
Remove qualified_name for Device

tetienne:
I just notice, the qualified name is within definition but not at the root of the device. You can check here: https://github.com/iMicknl/python-tahoma-api/blob/master/tests/devices.json


iMicknl (approved):
Good catch! Are you sure we don't use this anywhere?

tetienne:
Here in the library, we don’t use it. And within HA also, and it makes sense as the value is always None.

PR #191, opened 2021-09-26
Support SSO token and rewrite default server

### LINKED ISSUE #188: Add support for SSO token
Nexity requires a SSOToken for login, and we currently don't support this in the API.

However, this will add additional complexity since we add the user and pass to the constructor of the TahomaClient, and not just at the login().

iMicknl:
Fixes #188


iMicknl:
@tetienne would you be okay if I merge this one in? It would be great if we can abstract this more in the v1.0 release, but I think we need more major changes to support this better.

Agree?

PR #223, opened 2021-10-29
Add /setup endpoint

iMicknl:
- [x] Implement better caching (for `get_devices` and `get_gateways`)
- [x] Add tests (with multiple JSON files)

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -4,6 +4,9 @@ good-names=id,tb
 [LOGGING]
 logging-format-style=new
 
+[FORMAT]
+max-line-length=140
+
 [MESSAGES CONTROL]
 disable=
      logging-fstring-interpolation,
--- a/pyhoma/client.py
+++ b/pyhoma/client.py
@@ -247,7 +247,10 @@ class TahomaClient:
         response = await self.__get("setup")
         setup = Setup(**humps.decamelize(response))
 
+        # Cache response
         self.setup = setup
+        self.gateways = setup.gateways
+        self.devices = setup.devices
 
         return setup
 
@@ -266,7 +269,11 @@ class TahomaClient:
 
         response = await self.__get("setup/devices")
         devices = [Device(**d) for d in humps.decamelize(response)]
+
+        # Cache response
         self.devices = devices
+        if self.setup:
+            self.setup.devices = devices
 
         return devices
 
@@ -285,7 +292,11 @@ class TahomaClient:
 
         response = await self.__get("setup/gateways")
         gateways = [Gateway(**g) for g in humps.decamelize(response)]
+
+        # Cache response
         self.gateways = gateways
+        if self.setup:
+            self.setup.gateways = gateways
 
         return gateways
 
--- a/pyhoma/client.py
+++ b/pyhoma/client.py
@@ -47,6 +47,7 @@ from pyhoma.models import (
     OverkizServer,
     Place,
     Scenario,
+    Setup,
     State,
 )
 
@@ -59,6 +60,9 @@ async def relogin(invocation: dict[str, Any]) -> None:
 
 async def refresh_listener(invocation: dict[str, Any]) -> None:
     await invocation["args"][0].register_event_listener()
+
+
+# pylint: disable=too-many-instance-attributes
 
 
 class TahomaClient:
@@ -84,6 +88,7 @@ class TahomaClient:
         self.password = password
         self.server = server
 
+        self.setup: Setup | None = None
         self.devices: list[Device] = []
         self.gateways: list[Gateway] = []
         self.event_listener_id: str | None = None
@@ -225,6 +230,26 @@ class TahomaClient:
                 raise NexityServiceException("No Nexity SSO token provided.")
 
             return token["token"]
+
+    @backoff.on_exception(
+        backoff.expo,
+        (NotAuthenticatedException, ServerDisconnectedError),
+        max_tries=2,
+        on_backoff=relogin,
+    )
+    async def get_setup(self, refresh: bool = False) -> Setup:
+        """
+        Retrieve full setup
+        """
+        if self.setup and not refresh:
+            return self.setup
+
+        response = await self.__get("setup")
+        setup = Setup(**humps.decamelize(response))
+
+        self.setup = setup
+
+        return setup
 
     @backoff.on_exception(
         backoff.expo,
--- a/pyhoma/client.py
+++ b/pyhoma/client.py
@@ -239,7 +239,20 @@ class TahomaClient:
     )
     async def get_setup(self, refresh: bool = False) -> Setup:
         """
-        Retrieve full setup
+        Get all data about the connected user setup
+            -> gateways data (serial number, activation state, ...): <gateways/gateway>
+            -> setup location: <location>
+            -> house places (rooms and floors): <place>
+            -> setup devices: <devices>
+
+        A gateway may be in different modes (mode) regarding to the activated functions (functions).
+        A house may be composed of several floors and rooms. The house, floors and rooms are viewed as a place.
+        Devices in the house are grouped by type called uiClass. Each device has an associated widget.
+        The widget is used to control or to know the device state, whatever the device protocol (controllable): IO, RTS, X10, ... .
+        A device can be either an actuator (type=1) or a sensor (type=2).
+        Data of one or several devices can be also get by setting the device(s) url as request parameter.
+
+        Per-session rate-limit : 1 calls per 1d period for this particular operation (bulk-load)
         """
         if self.setup and not refresh:
             return self.setup
@@ -263,6 +276,7 @@ class TahomaClient:
     async def get_devices(self, refresh: bool = False) -> list[Device]:
         """
         List devices
+        Per-session rate-limit : 1 calls per 1d period for this particular operation (bulk-load)
         """
         if self.devices and not refresh:
             return self.devices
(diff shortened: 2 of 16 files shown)


iMicknl:
I would love if we can just do:
```python3
(devices, places, gateways) = await client.get_setup()
```

However, this is currently not supported:

```TypeError: cannot unpack non-iterable Setup object```.

Any ideas?

File: .pylintrc [hunk 1]
```diff
@@ -4,6 +4,9 @@ good-names=id,tb
 [LOGGING]
 logging-format-style=new
 
+[FORMAT]
+max-line-length=140
```
tetienne, line 8: That’s not compatible with black. Why?

pyhoma/client.py
```diff
@@ -84,6 +88,7 @@ def __init__(
         self.password = password
         self.server = server
 
+        self.setup: Setup | None = None
```
tetienne, line 91: FYI the | operator is only available since 3.10, and pyhoma require at least 3.9.

pyhoma/client.py
```diff
@@ -226,6 +231,42 @@ async def nexity_login(self) -> str:
...
+        if self.setup and not refresh:
+            return self.setup
+
+        response = await self.__get("setup")
+        setup = Setup(**humps.decamelize(response))
+
+        # Cache response
+        self.setup = setup
+        self.gateways = setup.gateways
+        self.devices = setup.devices
```
tetienne, line 266: Here we override the `self.devices` set within `get_devices`. It’s on purpose?

pyhoma/client.py
iMicknl: Ah, we use it in Home Assistant as well. Should I change it back, or just upgrade to 3.10?

However, doesn't `future annotations` solve this issue?

.pylintrc
iMicknl: Hmm, I wanted to have a bit more space for the documentation, since I am using some of the Overkiz documentation comments.

pyhoma/client.py
iMicknl: When we retrieve everything from /setup, we also get the devices. This way we save it to the cache, so when you call get_devices, it doesn't need to do an API call again for the same info.

tetienne: Perfect if it’s exactly the same response!

pyhoma/client.py
tetienne: Honnestly, I don’t know how to handle this properly.

iMicknl: By the way, this works already, look at the Python version of our test suite.

'''
from __future__ import annotations
'''

Does the trick

pyhoma/client.py
```diff
@@ -84,6 +88,7 @@ def __init__(
         self.password = password
         self.server = server
 
+        self.setup: Setup | None = None
         self.devices: list[Device] = []
         self.gateways: list[Gateway] = []
```
tetienne, line 93: I didn’t see this previously, but we should fix it. It’s not correct to assign [] as default value. See https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

pyhoma/models.py
```diff
@@ -22,13 +22,124 @@
...
-    return re.sub(r"\d+-", "****-", str(id))
+    return re.sub(r"(SETUP)?\d+-", "****-", str(id))
 
 
 def obfuscate_email(email: str | None) -> str:
     return re.sub(r"(.).*@.*(.\..*)", r"\1****@****\2", str(email))
 
 
+def mask(input: str | None) -> str:
+    return re.sub(r"[a-zA-Z0-9_.-]*", "*", str(input))
```
tetienne, line 33: So if None, it will return **** ?
```suggestion
    if input:
      return re.sub(r"[a-zA-Z0-9_.-]*", "*", str(input))
    return None
```

pyhoma/models.py
```diff
@@ -509,6 +619,73 @@ def __init__(
...
+
+    def __init__(
+        self,
+        *,
+        name: str,
+        source: str,
+        **_: Any,
+    ) -> None:
+        self.name = name
+        self.source = source
```
tetienne, line 635: I think it’s safe here to write. I don’t think Somfy will add new fields.

```suggestion
@attr.s(auto_attribs=True, slots=True, kw_only=True)
class Feature:
    name: str
    source: str
```

pyhoma/models.py
```diff
@@ -509,6 +619,73 @@ def __init__(
...
+
+
+@attr.s(auto_attribs=True, init=False, slots=True, kw_only=True)
+class ZoneItem:
```
tetienne, line 639: Same remark.

pyhoma/client.py
```diff
@@ -235,13 +276,18 @@ async def nexity_login(self) -> str:
...
         if self.devices and not refresh:
             return self.devices
 
         response = await self.__get("setup/devices")
         devices = [Device(**d) for d in humps.decamelize(response)]
+
+        # Cache response
         self.devices = devices
+        if self.setup:
+            self.setup.devices = devices
```
tetienne, line 290: So basically here, you don’t cache the devices, but sync with  the /setup entry point?

pyhoma/client.py
```diff
@@ -253,14 +299,19 @@ async def get_devices(self, refresh: bool = False) -> list[Device]:
...
 
         response = await self.__get("setup/gateways")
         gateways = [Gateway(**g) for g in humps.decamelize(response)]
+
+        # Cache response
         self.gateways = gateways
+        if self.setup:
+            self.setup.gateways = gateways
 
         return gateways
```
tetienne, line 317: What about:

```python
@property
def setup(self):
  if  self.setup:
    self.setup.gateways = self.gateways
    self.setup.devices = self.devices
  return setup
```

pyhoma/models.py
iMicknl: Good one!

pyhoma/client.py
iMicknl: What do you mean? What we do here is overwrite the /setup response with the latest devices response.

pyhoma/client.py
iMicknl: This won't work, since we need to have the latest cached information. It can happen that setup has newer information than the information in self.devices or self.gateways.

pyhoma/client.py
iMicknl: Should I set them to None, just like I did for self.setup?

tetienne: Yes have to do it and then set it to empty in the init

pyhoma/client.py
tetienne: OK it’s what I can sync :D

By the way, do we use this feature (cache) somewhere?

tests/fixtures/setup/setup_hi_kumo.json
```diff
@@ -0,0 +1,356 @@
+{
+    "creationTime": 1613674982000,
+    "lastUpdateTime": 1613674982000,
+    "id": "SETUP-1234-1234-6362",
+    "location": {
+        "creationTime": 1613674982000,
+        "lastUpdateTime": 1632527573000,
+        "city": "*",
```
tetienne, line 8: Would be great to have a "real" value to know what the data look like. Same remarks for all the * values.

iMicknl: Does it really matter? All * values are string values. I could replace the sensitive values with a placeholder, but since * is a string as well, I didn't bother.

tetienne: Hum OK.

pyhoma/client.py
iMicknl: @tetienne not really I think, thus I was also thinking to remove it... However, it doesn't have a lot of extra code and can help other people to make sure they don't call expensive endpoints all the time. :-)

tests/fixtures/setup/setup_hi_kumo.json
iMicknl: Would you like obfuscating more? So we keep the same amount of characters? I have the originals locally, so I can make changes to the regex. Currently I use the same regex as we use (in this PR) to obfuscate sensitive info in the logging.

pyhoma/models.py
iMicknl:
```
pyhoma/models.py:85: error: Argument "repr" has incompatible type "Callable[[Optional[str]], Optional[str]]"; expected "Union[bool, Callable[[Any], str]]"  [arg-type]
```

Apparently, this won't work. It seems that it will only be called when the value is not None.

pyhoma/client.py
iMicknl: I am not sure if I understand, since this is already the init method. Do I need to define these variables outside of my init as well? And then just set them in the init?

tetienne:
```
def __init__(self, my_array = None):
  if not my_array:
     my_array = []
```

iMicknl: But we are not setting them from the `init`? So I am not sure what to do here. Should I declare these fields outside of init and set them to empty in init?

```python3
    def __init__(
        self,
        username: str,
        password: str,
        server: OverkizServer,
        session: ClientSession = None,
    ) -> None:
        """
        Constructor

        :param username: the username for Tahomalink.com
        :param password: the password for Tahomalink.com
        :param server: OverkizServer
        :param session: optional ClientSession
        """

        self.username = username
        self.password = password
        self.server = server

        self.setup: Setup | None = None
        self.devices: list[Device] = []
        self.gateways: list[Gateway] = []
```

tetienne (approved):

PR #228, opened 2021-11-01
Add Overkiz State, Command, CommandParameter enums

iMicknl:


tetienne (commented):
This file starts to be too big. Can you add this in another module?

iMicknl:
> This file starts to be too big. Can you add this in another module?

I was looking into that indeed, how to more easily split this in multiple files. I would like to keep the import simple, instead of having to import from a specific file.

iMicknl:
@tetienne would you know if we can easily change the import to have it from `.enum` instead of `.enum.command`? Or should we just settle for new paths.

tetienne:
> @tetienne would you know if we can easily change the import to have it from `.enum` instead of `.enum.command`? Or should we just settle for new paths.

Perhaps adding the import within the `__init__.py` file. But, like this it does not hurt me.

pyhoma/enums/__init__.py
```diff
@@ -0,0 +1,8 @@
+# flake8: noqa: F401, F403
+
+from .command import *
+from .execution import *
+from .gateway import *
+from .general import *
+from .state import *
+from .ui import *
```
tetienne, line 8: So it was useful finally?

pyhoma/enums/ui.py
```diff
@@ -0,0 +1,396 @@
...
+    ZWAVE_DANFOSS_RSLINK = "ZWaveDanfossRSLink"
+    ZWAVE_DOOR_LOCK_CONFIGURATION = "ZWaveDoorLockConfiguration"
+    ZWAVE_FIBARO_ROLLER_SHUTTER_CONFIGURATION = "ZWaveFibaroRollerShutterConfiguration"
+    ZWAVE_HEATIT_THERMOSTAT_CONFIGURATION = "ZWaveHeatitThermostatConfiguration"
+    ZWAVE_NODON_CONFIGURATION = "ZWaveNodonConfiguration"
+    ZWAVE_QUBINO_CONFIGURATION = "ZWaveQubinoConfiguration"
+    ZWAVE_SEDEVICE_CONFIGURATION = "ZWaveSEDeviceConfiguration"
+    ZWAVE_TRANSCEIVER = "ZWaveTransceiver"
+    ZIGBEE_NETWORK = "ZigbeeNetwork"
+    ZIGBEE_STACK = "ZigbeeStack"
```
tetienne, line 396: That’s insane. How can they maintain all these widgets?

tetienne (approved):

pyhoma/enums/__init__.py
iMicknl: The only reason why I did this is to have everything as an import from `enums` (which is also non breaking :D).

pyhoma/enums/ui.py
iMicknl: I don't know haha, but they were already added to the enums. Didn't see that before :o.

PR #235, opened 2021-11-07
Add product type 7 (group)

iMicknl:
Coming from https://tahomalink.com/enduser-mobile-web/enduserAPI/reference/controllableTypes

```json
[
    {
        "id": 1,
        "name": "ACTUATOR",
        "label": "Actuator"
    },
    {
        "id": 2,
        "name": "SENSOR",
        "label": "Sensor"
    },
    {
        "id": 4,
        "name": "REMOTE_CONTROLLER",
        "label": "Remote Controller"
    },
    {
        "id": 5,
        "name": "PROTOCOL_GATEWAY",
        "label": "Protocol Gateway"
    },
    {
        "id": 3,
        "name": "VIDEO_SYSTEM",
        "label": "Video System"
    },
    {
        "id": 6,
        "name": "INFRASTRUCTURE_COMPONENT",
        "label": "Infrastructure Component"
    },
    {
        "id": 7,
        "name": "GROUP",
        "label": "Group Component"
    }
]
```

### DIFF
--- a/pyhoma/enums.py
+++ b/pyhoma/enums.py
@@ -6,6 +6,7 @@ _LOGGER = logging.getLogger(__name__)
 
 @unique
 class ProductType(IntEnum):
+    UNKNOWN = -1
     NONE = 0
     ACTUATOR = 1
     SENSOR = 2
@@ -13,6 +14,7 @@ class ProductType(IntEnum):
     CONTROLLABLE = 4
     GATEWAY = 5
     INFRASTRUCTURE_COMPONENT = 6
+    GROUP = 7
 
 
 @unique
--- a/pyhoma/enums.py
+++ b/pyhoma/enums.py
@@ -6,7 +6,6 @@ _LOGGER = logging.getLogger(__name__)
 
 @unique
 class ProductType(IntEnum):
-    UNKNOWN = -1
     NONE = 0
     ACTUATOR = 1
     SENSOR = 2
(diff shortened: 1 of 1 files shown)


tetienne (approved):

pyhoma/enums.py
```diff
@@ -6,13 +6,15 @@
 
 @unique
 class ProductType(IntEnum):
+    UNKNOWN = -1
```
tetienne, line 9: You miss the missing method

tetienne (changes requested):

pyhoma/enums.py
iMicknl: Oops, I decided to not add the method since this one is pretty stable and shouldn't change much.

But I forgot to remove the unknown enum. Would you agree that adding the extra logic is overkill? (Since we don't do it for all enums).

tetienne: Yes it will be overkill indeed here.

pyhoma/enums.py
```diff
@@ -13,6 +13,7 @@ class ProductType(IntEnum):
     CONTROLLABLE = 4
     GATEWAY = 5
     INFRASTRUCTURE_COMPONENT = 6
+    GROUP = 7
```
tetienne, line 16: Wondering what it can be

PR #275, opened 2021-12-19
Rebrand to OverKiz

iMicknl:
- [x] Release a new pyhoma version that mentions deprecation and move to overkiz + link
- [x] Rename GitHub repo to https://github.com/iMicknl/python-overkiz-api (should create a redirect)
- [ ] Release 1.0.0 version of pyoverkiz


.python-version
```diff
@@ -1 +1 @@
-3.7.8
```
tetienne, line 1: HA is not using Python 3.8?

iMicknl: Good one, let's make it 3.8.

tetienne (approved):

PR #278, opened 2021-12-21
Add deprecation notice for pyhoma

iMicknl:


vlebourl (approved):

README.md
```diff
@@ -1,3 +1,7 @@
+# Deprecation Notice
+**`pyhoma` is deprecated. Please use `pyoverkiz` 1.0 which is backwards compatible. See https://github.com/iMicknl/python-overkiz-api.**
```
tetienne, line 2: Correct me if I’m wrong, but Github will automatically redirect from python-tahoma-api to python-overkiz-api repository. So I don’t think the last part is required.

iMicknl: I added this for pypi, so we link them to our GitHub from there as well.

tetienne (approved):

PR #288, opened 2021-12-29
Revert "Bump tox from 3.24.4 to 3.24.5"

iMicknl:
Reverts iMicknl/python-overkiz-api#287


tetienne:
Apparently the issue was not from Tox

PR #293, opened 2022-01-03
Be more strict with Mypy

tetienne:


iMicknl:
Can you rebase on main to fix CI?

pyoverkiz/models.py
```diff
@@ -267,9 +267,15 @@ def __len__(self) -> int:
 class State:
     name: str
     type: DataType
-    value: str | None = None
+    value: None | int | float | str | bool
```
iMicknl, line 270: @tetienne I think this was the mistake that messed up the core PR...

pyoverkiz/client.py
```diff
@@ -196,13 +205,9 @@ async def nexity_login(self) -> str:
         loop = asyncio.get_event_loop()
 
         # Request access token
-        client = await loop.run_in_executor(
-            None,
-            lambda: boto3.client(
-                "cognito-idp", config=Config(region_name=NEXITY_COGNITO_REGION)
-            ),
+        client = boto3.client(
```
iMicknl, line 208: @tetienne why did this one get removed? This is currently breaking Nexity login.

tetienne: There was a typing error, and cannot found someone using our syntax, while this one yes :S

iMicknl: I resolved it in a new PR :)

PR #296, opened 2022-01-05
Add Siren RTD to setup fixtures / tests

iMicknl:
Good one to have in our collection, for future tests. (since I broke this one while adding the new Siren entity).

I added my masking file to the repo, so that others can add setup.json files as well and mask them accordingly.


utils/mask_fixtures.py
```diff
@@ -0,0 +1,78 @@
+from __future__ import annotations
+from collections import abc
+
+import re
+import json
+import os
+import glob
+import humps
+
+# list_not_to_mask = ["id"]
```
tetienne, line 10: ?

tetienne (approved):

utils/mask_fixtures.py
iMicknl: I can remove that, I don't use it.

PR #297, opened 2022-01-05
Fix CI (try 2)

iMicknl:


pyoverkiz/client.py
```diff
@@ -203,7 +203,7 @@ async def nexity_login(self) -> str:
             ),
         )
 
-        aws = AWSSRP(
+        aws = WarrantLite(
```
tetienne, line 206: That’s the fix?

iMicknl: It seems that both packages are pretty outdated, but that we could use the lite variant as well.

PR #300, opened 2022-01-05
Move on_backoff call to class method

iMicknl:
I didn't like the previous code style to be honest, quite a vague method to reference `self`. I was not able to try this, however according to the docs calling a class method should work as well.


tetienne:
Looks better indeed. You can try your code throwing yourself an error in a method. You will see if the backoff method is called or not.

iMicknl:
Closed in favor of #510

PR #302, opened 2022-01-06
Update device model

iMicknl:


pyoverkiz/models.py
```diff
@@ -151,8 +153,8 @@ class Device:
     controllable_name: str
     definition: Definition
     data_properties: list[dict[str, Any]] | None = None
-    widget: str | None = None
```
iMicknl, line 154: Do we know why these were optional? I think every device has a widget and uiclass set.

pyoverkiz/models.py
```diff
@@ -185,8 +187,8 @@ def __init__(
         self.controllable_name = controllable_name
         self.states = States(states)
         self.data_properties = data_properties
-        self.widget = widget
-        self.ui_class = ui_class
+        self.widget = UIWidget(widget)
+        self.ui_class = UIClass(ui_class)
```
tetienne, line 191: We should probably add the missing methods to these enums in case Somfy add a new widget.

iMicknl: Done

tetienne (approved):

PR #305, opened 2022-01-06
Move casting to event listener and add OverkizStateType

iMicknl:


pyoverkiz/types.py
```diff
@@ -0,0 +1,18 @@
...
+OverkizStateType = Union[str, int, float, bool, Dict[Any, Any], List[Any], None]
+
+DATA_TYPE_TO_PYTHON: dict[DataType, Callable[[Any], OverkizStateType]] = {
+    DataType.INTEGER: int,
+    DataType.DATE: int,
+    DataType.STRING: str,
+    DataType.FLOAT: float,
+    DataType.BOOLEAN: bool,
+    DataType.JSON_ARRAY: json.loads,
+    DataType.JSON_OBJECT: json.loads,
```
tetienne, line 17: It will avoid the special case in your init.
```suggestion
    DataType.JSON_OBJECT: json.loads,
    DataType.NONE: lambda x: x
```

pyoverkiz/types.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import annotations
+
+import json
+from typing import Any, Callable, Dict, List, Union
+
+from pyoverkiz.enums import DataType
+
+OverkizStateType = Union[str, int, float, bool, Dict[Any, Any], List[Any], None]
```
tetienne, line 8:
```suggestion
OVERKIZ_STATE_TYPE = Union[str, int, float, bool, Dict[Any, Any], List[Any], None]
```

pyoverkiz/models.py
```diff
@@ -269,20 +270,38 @@ def __len__(self) -> int:
...
+    def __init__(
+        self,
+        name: str,
+        type: int,
+        value: str | None = None,
+        **_: Any,
+    ):
+        super().__init__(name, type, value, **_)
+
+        if self.type == DataType.NONE:
```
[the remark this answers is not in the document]
iMicknl: No clue, I have never seen this to be honest. This ENUM came from Somfy directly.

pyoverkiz/types.py
iMicknl: It seems that for types (so no official constant?), this is not enforced. At least not in Home Assistant. https://github.com/home-assistant/core/blob/dev/homeassistant/helpers/typing.py

Perhaps we should just call it StateType and import it as OverkizStateType when we use it in Home Assistant. Would that be okay?

tetienne: It’s ok for me.

pyoverkiz/types.py
iMicknl:
```
pyoverkiz/types.py:11: error: Returning Any from function declared to return "Union[str, int, float, bool, Dict[Any, Any], List[Any], None]"  [no-any-return]
Found 1 error in 1 file (checked 2 source files)
```

iMicknl: It seems that this is not possible, I will need to have a look.

tetienne: I guess we have to type the lambda

iMicknl: I just did :D

vlebourl (approved):

tetienne (approved):

PR #310, opened 2022-01-11
Add setOperatingMode + params

egguy:
Add new command and params, missing in the Somfy Heating Temperature Interface entity.


tetienne (changes requested):
Can you please sort alphabetically these new constants?

iMicknl (approved):

PR #312, opened 2022-01-12
Make last update time optional + add tests

iMicknl:
Fixes https://github.com/iMicknl/ha-tahoma/issues/706


tetienne (approved):
Do you know why for this user, the field was not returned? Their is something special?

iMicknl:
> Do you know why for this user, the field was not returned? Their is something special?

No... Only that he migrated to the TaHoma V2 hub, from a V1 hub.

PR #315, opened 2022-01-13
Add first implementation of Somfy OAuth

iMicknl:
This currently works but will relogin via backoff every hour (see expires). Eventually best would be to see if we can improve the Somfy, Cozytouch and Nexity login code since none of them handle the tokens currently.

```json
{
	"access_token": "",
	"expires_in": 3600,
	"token_type": "bearer",
	"scope": "user.basic api.full oa.site oa.subscription oa.user oa.device oa.devicedefinition level.0",
	"refresh_token": ""
}
```

### DIFF
--- a/pyoverkiz/client.py
+++ b/pyoverkiz/client.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 import urllib.parse
 from json import JSONDecodeError
 from types import TracebackType
@@ -56,6 +57,7 @@ from pyoverkiz.models import (
     State,
 )
 
+_LOGGER = logging.getLogger(__name__)
 JSON = Union[Dict[str, Any], List[Dict[str, Any]]]
 
 
@@ -139,6 +141,7 @@ class OverkizClient:
         # Somfy TaHoma authentication using access_token
         if self.server == SUPPORTED_SERVERS["somfy_europe"]:
             access_token = await self.somfy_tahoma_login()
+            _LOGGER.debug("Using Somfy OAuth login")
             payload = {"accessToken": access_token}
 
         # CozyTouch authentication using jwt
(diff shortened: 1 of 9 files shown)


tetienne (commented):
It would be nice to start implementing a factory, it will become hard to maintain.

pyoverkiz/const.py
```diff
@@ -72,3 +76,9 @@
         configuration_url=None,
     ),
 }
+
+CLIENTS: dict[str, OverkizClient] = {
```
iMicknl, line 80: Would be good to make OverkizClient part of the OverkizServer implementation? To have everything in one place, and even allow developers to pass their own OverkizServer object, with a custom Client.

pyoverkiz/clients/somfy_client.py
```diff
@@ -0,0 +1,68 @@
...
+from pyoverkiz.client import OverkizClient
+from pyoverkiz.const import SOMFY_API, SOMFY_CLIENT_ID, SOMFY_CLIENT_SECRET
+from pyoverkiz.exceptions import (
+    BadCredentialsException,
+    SomfyBadCredentialsException,
+    SomfyServiceException,
+)
+
+
+class SomfyEuropeClient(OverkizClient):
```
iMicknl, line 11: Is this the class where we will store the tokens and handle the refresh?

pyoverkiz/client.py
```diff
@@ -82,25 +64,26 @@ class OverkizClient:
     event_listener_id: str | None
     session: ClientSession
 
+    @staticmethod
+    def get_client(server: OverkizServer) -> OverkizClient:
+        return CLIENTS.get(server, SimpleClient)
+
     def __init__(
         self,
         username: str,
```
iMicknl, line 73: Eventually we also need to think how we handle this, since local auth doesn't require a username / password.

pyoverkiz/clients/somfy_client.py
tetienne: Yes that's the idea.

pyoverkiz/client.py
tetienne: That's good to know.

pyoverkiz/const.py
tetienne: Done with e3f2461

tetienne:
Can be closed I guess?

PR #316, opened 2022-01-13
Add Rexel energeasy connect gateway type

egguy:
I don't know if it's useful, but the subType is 0, should I add this ?


iMicknl (approved):
Thanks! And subType 0 is already in the list, this doesn't need to be mapped.

PR #324, opened 2022-01-18
Add diagnostics endpoint

iMicknl:
Decided to wrap it in a new function call and not extend `get_setup()` since that will mess up the typing a lot, and now we can directly mask the data.

```
        setup = await client.get_diagnostic_data()
```

### DIFF
--- a/pyoverkiz/client.py
+++ b/pyoverkiz/client.py
@@ -5,7 +5,7 @@ import asyncio
 import urllib.parse
 from json import JSONDecodeError
 from types import TracebackType
-from typing import Any, Dict, List, Union, cast
+from typing import Any, cast
 
 import backoff
 import boto3
@@ -50,8 +50,8 @@ from pyoverkiz.models import (
     Setup,
     State,
 )
-
-JSON = Union[Dict[str, Any], List[Dict[str, Any]]]
+from pyoverkiz.types import JSON
+from pyoverkiz.utils import mask_json
 
 
 async def relogin(invocation: dict[str, Any]) -> None:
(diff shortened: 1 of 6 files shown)


pyoverkiz/utils.py
```diff
@@ -0,0 +1,64 @@
+"""Utils for Overkiz client"""
```
tetienne, line 1: Not a big fan of `utils` name. This is too generic. I would suggest something like `obfuscate.py` or `mask.py`.

pyoverkiz/utils.py
```diff
@@ -0,0 +1,64 @@
...
+    """Mask id"""
+    return re.sub(r"(SETUP)?\d+-", "****-", str(id))
+
+
+def obfuscate_email(email: str | None) -> str:
+    """Mask email"""
+    return re.sub(r"(.).*@.*(.\..*)", r"\1****@****\2", str(email))
+
+
+def mask(input: str) -> str:
```
tetienne, line 19: `mask` vs `obfuscate`, we have to choose.

iMicknl: For now I just ported the names from our current `models.py`. However happy to rename and list as breaking change.

pyoverkiz/utils.py
iMicknl: Fine by me as well, let's change it to what we decide the function name to be.

pyoverkiz/utils.py
```diff
@@ -0,0 +1,64 @@
...
+    """Mask string"""
+    return re.sub(r"[a-zA-Z0-9_.-]*", "*", str(input))
+
+
+# pylint: disable=too-many-nested-blocks
+def mask_json(data: dict) -> JSON:
```
tetienne, line 25:
```suggestion
def mask_sensitive_data(data: dict) -> JSON:
```

tetienne: dict is also missing type.

pyoverkiz/utils.py
```diff
@@ -0,0 +1,64 @@
...
+        if isinstance(value, dict):
+            mask_json(value)
+        elif isinstance(value, list):
+            for val in value:
+                if isinstance(val, str):
+                    pass
+                elif isinstance(val, list):
+                    pass
+                else:
+                    mask_json(val)
```
tetienne, line 62:
```suggestion
                if isinstance(val, str):
                    continue
                elif isinstance(val, list):
                    continue
                else:
                    mask_json(val)
```

pyoverkiz/client.py
```diff
@@ -272,6 +273,27 @@ async def get_setup(self, refresh: bool = False) -> Setup:
...
+            -> setup location: <location>
+            -> house places (rooms and floors): <place>
+            -> setup devices: <devices>
+
+        This data will be masked to not return any confidential or PII data.
+        """
+        response = await self.__get("setup")
+        setup = obfuscate_sensitive_data(response)
+
+        return setup
```
tetienne, line 295:
```suggestion

        return obfuscate_sensitive_data(response)
```

tetienne (approved):

PR #330, opened 2022-01-21
Add more specific state value getters

tetienne:
- Add more accurrate state getter
- Add more  specific value getter

### DIFF
--- a/pyoverkiz/models.py
+++ b/pyoverkiz/models.py
@@ -274,6 +274,22 @@ class State:
         self.value = value
         self.type = DataType(type)
 
+    @property
+    def value_as_int(self) -> int:
+        return int(self.value)
+
+    @property
+    def value_as_float(self) -> float:
+        return float(self.value)
+
+    @property
+    def value_as_bool(self) -> bool:
+        return bool(self.value)
+
+    @property
+    def value_as_str(self) -> str:
+        return str(self.value)
+
 
 @define(init=False, kw_only=True)
 class EventState(State):
--- a/pyoverkiz/models.py
+++ b/pyoverkiz/models.py
@@ -295,7 +295,7 @@ class State:
         if self.type == DataType.NONE:
             return None
         if self.type == DataType.BOOLEAN:
-            return bool(self.value)
+            return cast(bool, self.value)
         raise TypeError(f"{self.name} is not a boolean")
 
     @property
@@ -303,7 +303,7 @@ class State:
         if self.type == DataType.NONE:
             return None
         if self.type == DataType.STRING:
-            return str(self.value)
+            return cast(str, self.value)
         raise TypeError(f"{self.name} is not a string")
 
 
--- a/pyoverkiz/models.py
+++ b/pyoverkiz/models.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, Iterator
+from typing import Any, Iterator, cast
 
 from attr import define, field
 
@@ -276,19 +276,27 @@ class State:
 
     @property
     def value_as_int(self) -> int:
-        return int(self.value)
+        if self.type == DataType.INTEGER:
+            return cast(int, self.value)
+        raise TypeError(f"{self.name} is not an integer")
 
     @property
     def value_as_float(self) -> float:
-        return float(self.value)
+        if self.type == DataType.FLOAT:
+            return cast(float, self.value)
+        raise TypeError(f"{self.name} is not a float")
 
     @property
     def value_as_bool(self) -> bool:
-        return bool(self.value)
+        if self.type == DataType.BOOLEAN:
+            return bool(self.value)
+        raise TypeError(f"{self.name} is not a boolean")
 
     @property
     def value_as_str(self) -> str:
-        return str(self.value)
+        if self.type == DataType.STRING:
+            return str(self.value)
+        raise TypeError(f"{self.name} is not a string")
 
 
 @define(init=False, kw_only=True)
--- a/pyoverkiz/models.py
+++ b/pyoverkiz/models.py
@@ -311,7 +311,7 @@ class State:
         if self.type == DataType.NONE:
             return None
         if self.type == DataType.JSON_OBJECT:
-            return cast(dict[str, Any], self.value)
+            return cast(dict, self.value)
         raise TypeError(f"{self.name} is not a JSON object")
 
     @property
@@ -319,7 +319,7 @@ class State:
         if self.type == DataType.NONE:
             return None
         if self.type == DataType.JSON_ARRAY:
-            return cast(list[Any], self.value)
+            return cast(list, self.value)
         raise TypeError(f"{self.name} is not an array")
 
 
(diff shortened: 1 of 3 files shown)


pyoverkiz/models.py
```diff
@@ -271,6 +271,30 @@ def __init__(
         self.value = value
         self.type = DataType(type)
 
+    @property
+    def value_as_int(self) -> int:
```
epenet, line 275: Does this take into account None values? Or are they not possible?
```suggestion
    def value_as_int(self) -> int | None:
```

tetienne: I don’t remember a case where the API returned a state without any value. @iMicknl ? There was not also a DataType.NONE?

epenet: I saw a few None checks inside the HA code, but I didn't look too deep.
I thought maybe something like this was safer
```python
        if self.type == DataType.NONE:
            return None
        if self.type == DataType.INTEGER:
            return cast(int, self.value)
        raise TypeError(f"{self.name} is not an integer")
```

iMicknl: self.value can be None, this is the case when `DataType.None`
https://github.com/iMicknl/python-overkiz-api/blob/4d9745f9c272f96d05a60a9e49e577894206648f/pyoverkiz/models.py#L267

We also have an example of that in our test fixtures.

tetienne: @iMicknl `DataType.None` does not exist. I double check within TahomaLink website source, and they still haven’t:

```
        switch (t) {
          case DataType.INTEGER:
            return Number.parseInt(e);
          case DataType.FLOAT:
            return Number.parseFloat(e);
          case DataType.BOOLEAN:
            return r(e);
          case DataType.STRING:
          case DataType.BLOB:
            return e;
          case DataType.DATE:
          case DataType.JSON_OBJECT:
          case DataType.JSON_ARRAY:
            return "string" == typeof e ? JSON.parse(e) : e;
        }
```

iMicknl: @tetienne see https://github.com/iMicknl/python-overkiz-api/blob/24b9bd708f6d0dd7b2f1846dec15a45e61e9c293/tests/events.json#L30-L34 for an example of DataType 0. (value is missing).

tetienne: Thx I missed this one:
https://github.com/iMicknl/python-overkiz-api/blob/24b9bd708f6d0dd7b2f1846dec15a45e61e9c293/pyoverkiz/enums/general.py#L21

So we are fine, Overkize indicate if there is a value or not. Meaning if we DataType.INTEGER (or whatever), we are sure to get a value.

iMicknl: My only concern is that it seems that this can change. `core:ConsumptionTariff4State` is an int normally, however for this devcice the state is empty. So shouldn't we now always need to check if the state is not empty?

tetienne: I see what you mean. So you want I returned None in such case as suggested by @epenet?

iMicknl: I think we just need to see how this current change works with Overkiz in core (just in a dev branch) and than make changes accordingly.

tetienne: OK I’ve implemented @epenet suggestion. So it will force the client to check the returned value. 

About the method names, it’s ok? I hesitated between `int_value` and `value_as_int`.

tetienne:
@iMicknl What do you think?

pyoverkiz/models.py
```diff
@@ -271,6 +271,38 @@ def __init__(
...
+        if self.type == DataType.FLOAT:
+            return cast(float, self.value)
+        raise TypeError(f"{self.name} is not a float")
+
+    @property
+    def value_as_bool(self) -> bool | None:
+        if self.type == DataType.NONE:
+            return None
+        if self.type == DataType.BOOLEAN:
+            return bool(self.value)
```
iMicknl, line 295: Why cast it everywhere and not here?

pyoverkiz/models.py
```diff
@@ -271,6 +271,38 @@ def __init__(
...
+        raise TypeError(f"{self.name} is not a boolean")
+
+    @property
+    def value_as_str(self) -> str | None:
+        if self.type == DataType.NONE:
+            return None
+        if self.type == DataType.STRING:
+            return str(self.value)
```
iMicknl, line 303: Why cast it (almost) everywhere and not here?

pyoverkiz/models.py
```diff
@@ -271,6 +271,38 @@ def __init__(
...
+        raise TypeError(f"{self.name} is not a string")
+
```
iMicknl, line 306: Perhaps add value as dict as well?

pyoverkiz/models.py
tetienne: The issue is where I cast and not here.

It’s because of the returned type of value:

```
pyoverkiz/models.py:282: error: Argument 1 to "int" has incompatible type "Union[str, float, Dict[Any, Any], List[Any], None]"; expected "Union[str, bytes, SupportsInt, SupportsIndex, SupportsTrunc]"  [arg-type]

pyoverkiz/models.py:290: error: Argument 1 to "float" has incompatible type "Union[str, float, Dict[Any, Any], List[Any], None]"; expected "Union[SupportsFloat, SupportsIndex, str, bytes]"  [arg-type]
```

iMicknl: But shouldn't we just use cast() everywhere? (since we should not need to do a implicit cast via int() / str() again?)

tetienne: We won't do any additional cast after this. I don't get your concern.

epenet: `cast()` is free and only used by type checking.
`bool()` has a (very small) cost.
If you are sure that it is already a boolean, then `cast` is cheaper.

tetienne: Thx @epenet I was not sure about the more pythonic way to do it.

pyoverkiz/models.py
tetienne: Well, what can of type should I return?

tetienne: Done

pyoverkiz/models.py
```diff
@@ -274,6 +274,54 @@ def __init__(
...
+        if self.type == DataType.JSON_OBJECT:
+            return cast(dict[str, Any], self.value)
+        raise TypeError(f"{self.name} is not a JSON object")
+
+    @property
+    def value_as_list(self) -> list[Any] | None:
+        if self.type == DataType.NONE:
+            return None
+        if self.type == DataType.JSON_ARRAY:
+            return cast(list[Any], self.value)
```
tetienne, line 322: I was expecting to cast to `list[dict[str, Any]]`, but in some cases we have a list of string.

iMicknl: I think it is mainly a list of strings indeed.

pyoverkiz/models.py
iMicknl: Awesome

iMicknl (commented):
I am still in doubt if we should add the check for None... Could we try to implement this somewhere before we merge? So we know if this optional None is not causing any big issues?

iMicknl (approved):

PR #331, opened 2022-01-21
Add devcontainer 

iMicknl:
Not perfect yet, however would help me a lot to speed up development :-).


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,42 @@
...
+        "python.linting.pylintEnabled": true,
+        "python.formatting.autopep8Path": "/usr/local/py-utils/bin/autopep8",
+        "python.formatting.blackPath": "/usr/local/py-utils/bin/black",
+        "python.formatting.yapfPath": "/usr/local/py-utils/bin/yapf",
+        "python.linting.banditPath": "/usr/local/py-utils/bin/bandit",
+        "python.linting.flake8Path": "/usr/local/py-utils/bin/flake8",
+        "python.linting.mypyPath": "/usr/local/py-utils/bin/mypy",
+        "python.linting.pycodestylePath": "/usr/local/py-utils/bin/pycodestyle",
+        "python.linting.pydocstylePath": "/usr/local/py-utils/bin/pydocstyle",
+        "python.linting.pylintPath": "/usr/local/py-utils/bin/pylint",
```
tetienne, line 23: We should move these parameters into the settings.json file instead, and targetting .venv/bin folder. So even if we don’t use the dev container Code will be correctly setup.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,42 @@
...
+        "terminal.integrated.profiles.linux": {
+            "zsh": {
+                "path": "/usr/bin/zsh"
+            }
+        },
+        "terminal.integrated.defaultProfile.linux": "zsh"
+    },
+    "extensions": [
+        "ms-python.python",
+        "ms-python.vscode-pylance",
```
tetienne, line 33: This one is built-in the python extension

.devcontainer/devcontainer.json
iMicknl: You don't use a venv inside a container, thus that wouldn't work? However I copied this from the Home Assistant devcontainer and I am not sure it this all works correctly.

PR #333, opened 2022-01-23
Move to Somfy OAuth as access method for Somfy Europe

iMicknl:
Not pretty code, however we need to see if this will fix the issues.

### DIFF
--- a/pyoverkiz/client.py
+++ b/pyoverkiz/client.py
@@ -242,10 +242,6 @@ class OverkizClient:
             self._expires_in = datetime.datetime.now() + datetime.timedelta(
                 seconds=int(token["expires_in"]) - 5
             )
-            print(token)
-
-            if self.event_listener_id:
-                await self.register_event_listener()
 
             return self._access_token
 
@@ -608,13 +604,7 @@ class OverkizClient:
         headers = {}
 
         if self.server == SUPPORTED_SERVERS["somfy_europe"]:
-            if (
-                self._expires_in
-                and self._refresh_token
-                and self._expires_in <= datetime.datetime.now()
-            ):
-                await self.somfy_tahoma_refresh_token(self._refresh_token)
-
+            await self._refresh_somfy_tahoma_token_if_expired()
             headers["Authorization"] = f"Bearer {self._access_token}"
 
         async with self.session.get(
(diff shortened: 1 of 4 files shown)


pyoverkiz/client.py
```diff
@@ -155,6 +173,78 @@ async def login(
...
+            token = await response.json()
+            # { "message": "error.invalid.grant", "data": [], "uid": "xxx" }
+            if "message" in token and token["message"] == "error.invalid.grant":
+                raise SomfyBadCredentialsException(token["message"])
+
+            if "access_token" not in token:
+                raise SomfyServiceException("No Somfy access token provided.")
+
+            self._access_token = str(token["access_token"])
+            self._refresh_token = str(token["refresh_token"])
```
tetienne, line 241: They are not already a string?

pyoverkiz/client.py
```diff
@@ -155,6 +173,78 @@ async def login(
...
+            self._expires_in = datetime.datetime.now() + datetime.timedelta(
+                seconds=int(token["expires_in"]) - 5
```
tetienne, line 243: It’s not already an int into the response?

pyoverkiz/client.py
```diff
@@ -508,25 +601,45 @@ async def execute_scenario(self, oid: str) -> str:
...
             return await response.json()
 
     async def __delete(self, path: str) -> None:
         """Make a DELETE request to the OverKiz API"""
-        async with self.session.delete(f"{self.server.endpoint}{path}") as response:
+        headers = {}
+
+        if self.server == SUPPORTED_SERVERS["somfy_europe"]:
+            await self._refresh_somfy_tahoma_token_if_expired()
+            headers["Authorization"] = f"Bearer {self._access_token}"
```
tetienne, line 638: Can put somewhere else later to avoid dup code.

tetienne (approved):
Just think about removing all the print statements, and for a 1st version,it’s ok. 

We will be able to clean it later.

pyoverkiz/client.py
iMicknl: I thought about that one, and pass the headers to that function, however for now I found this clear enough.

Would this be ok for this current fix?

pyoverkiz/client.py
iMicknl: I think so... Would need to double check. Same for the int.

tetienne (approved):

PR #336, opened 2022-01-24
Add protocol property to Device model

tetienne:


tetienne:
I don’t get why I haven’t this duplicate code error locally :o

iMicknl:
> I don’t get why I haven’t this duplicate code error locally :o

Because you don't run Python 3.8? However I have seen this error before, but I don't recall where.

image

tetienne:
I ran the `tox` command too on my machine, and also `poetry run pre-commit run -a`. You can reproduce?

tetienne:
@iMicknl Fixed :D

vlebourl (approved):

iMicknl (changes requested):
https://github.com/iMicknl/python-overkiz-api/blob/main/pyoverkiz/enums/__init__.py

Could you perhaps add this one to init as well? Most of the enums we just import from `pyoverkiz.enums`.

tetienne:
> https://github.com/iMicknl/python-overkiz-api/blob/main/pyoverkiz/enums/__init__.py
> 
> Could you perhaps add this one to init as well? Most of the enums we just import from `pyoverkiz.enums`.

Done

iMicknl (approved):

PR #340, opened 2022-01-25
Obfuscate label in logging

iMicknl:
Label is used for the user friendly name as well, we do redact this in our diagnostic logging, but not in our debug logging yet.

### DIFF
--- a/pyoverkiz/models.py
+++ b/pyoverkiz/models.py
@@ -138,7 +138,7 @@ class Device:
     attributes: States
     available: bool
     enabled: bool
-    label: str = field(repr=obfuscate_string, default=None)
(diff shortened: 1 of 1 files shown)


vlebourl (approved):

pyoverkiz/models.py
```diff
@@ -447,7 +447,7 @@ def __init__(
 
 @define(init=False, kw_only=True)
 class Scenario:
-    label: str
+    label: str = field(repr=obfuscate_string, default=None)
```
tetienne, line 450: What set a default value?

iMicknl: I copied that, will remove it since str is required.

PR #346, opened 2022-01-27
Refresh token before using listener id

tetienne:
.

### DIFF
--- a/pyoverkiz/client.py
+++ b/pyoverkiz/client.py
@@ -211,10 +211,13 @@ class OverkizClient:
 
             return self._access_token
 
-    async def somfy_tahoma_refresh_token(self, refresh_token: str) -> str:
+    async def refresh_token(self) -> None:
         """
-        To get a new access token, the refresh token is needed. The refresh_token will be valid 14 days.
+        Update the access and the refresh token. The refresh token will be valid 14 days.
         """
+        if self.server != SUPPORTED_SERVERS["somfy_europe"]:
+            return
+
         # &grant_type=refresh_token&refresh_token=REFRESH_TOKEN
         # Request access token
         async with self.session.post(
@@ -222,7 +225,7 @@ class OverkizClient:
             data=FormData(
                 {
                     "grant_type": "refresh_token",
-                    "refresh_token": refresh_token,
+                    "refresh_token": self._refresh_token,
                     "client_id": SOMFY_CLIENT_ID,
                     "client_secret": SOMFY_CLIENT_SECRET,
                 }
@@ -244,8 +247,6 @@ class OverkizClient:
             self._expires_in = datetime.datetime.now() + datetime.timedelta(
                 seconds=token["expires_in"] - 5
             )
-
-            return self._access_token
 
     async def cozytouch_login(self) -> str:
         """
@@ -724,15 +725,12 @@ class OverkizClient:
 
     async def _refresh_token_if_expired(self) -> None:
         """Check if token is expired and request a new one."""
-        if self.server != SUPPORTED_SERVERS["somfy_europe"]:
-            return
-
         if (
             self._expires_in
             and self._refresh_token
             and self._expires_in <= datetime.datetime.now()
         ):
-            await self.somfy_tahoma_refresh_token(self._refresh_token)
+            await self.refresh_token()
 
             if self.event_listener_id:
                 await self.register_event_listener()
--- a/pyoverkiz/client.py
+++ b/pyoverkiz/client.py
@@ -128,7 +128,6 @@ class OverkizClient:
     async def close(self) -> None:
         """Close the session."""
         if self.event_listener_id:
-            await self._refresh_token_if_expired()
             await self.unregister_event_listener()
 
         await self.session.close()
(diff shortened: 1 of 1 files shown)


pyoverkiz/client.py
```diff
@@ -244,8 +248,6 @@ async def somfy_tahoma_refresh_token(self, refresh_token: str) -> str:
                 seconds=token["expires_in"] - 5
             )
 
-            return self._access_token
```
tetienne, line 247: It was not used, as we can directly use `self._access_token`

pyoverkiz/client.py
```diff
@@ -210,18 +211,21 @@ async def somfy_tahoma_get_access_token(self) -> str:
 
             return self._access_token
 
-    async def somfy_tahoma_refresh_token(self, refresh_token: str) -> str:
+    async def refresh_token(self) -> None:
```
tetienne, line 214: We don’t need to give in parameter the refresh_token, we can directly use the inner one.

pyoverkiz/client.py
```diff
@@ -128,6 +128,7 @@ async def __aexit__(
     async def close(self) -> None:
         """Close the session."""
         if self.event_listener_id:
+            await self._refresh_token_if_expired()
```
iMicknl, line 131: Why is this required? Since __post calls this function as well?

tetienne: Indeed, this is a mistake.

pyoverkiz/client.py
iMicknl: Indeed! Do we need to check if we have a refresh token, or can this race condition never happen.

pyoverkiz/client.py
```diff
@@ -519,6 +521,7 @@ async def fetch_events(self) -> list[Event]:
         Per-session rate-limit : 1 calls per 1 SECONDS period for this particular
         operation (polling)
         """
+        await self._refresh_token_if_expired()
```
iMicknl, line 523: Why is this required here? Since we call post a line later and in post we also call this function?

pyoverkiz/client.py
```diff
@@ -629,8 +633,8 @@ async def __get(self, path: str) -> Any:
         """Make a GET request to the OverKiz API"""
         headers = {}
 
-        if self.server == SUPPORTED_SERVERS["somfy_europe"]:
-            await self._refresh_somfy_tahoma_token_if_expired()
+        await self._refresh_token_if_expired()
```
iMicknl, line 635: Just a double check, here we call this before we check if there is an access token. In post we call this after we checked if we have an access token.

pyoverkiz/client.py
iMicknl: Why the rename? I didn't like the previous one as well, however this only applies to Somfy TaHoma (Europe).

pyoverkiz/client.py
tetienne: In our code, we have already a check: https://github.com/iMicknl/python-overkiz-api/blob/0c45f2e6a3ae20b7af77ec8cd976a393a2a3f780/pyoverkiz/client.py#L729

but not directly in this method,  I can raise an error in such case. What do you think?

pyoverkiz/client.py
tetienne: That’s because in the `__post`, I don’t want this method to be call for the `login` endpoint. It’s not about the access_token.

pyoverkiz/client.py
tetienne: It’s with the idea to have something generic, and also avoid to test each time the server type before calling this method.

pyoverkiz/client.py
iMicknl: Would be good to raise an error / exception (or call the other function), just to be sure.

iMicknl:
@tetienne will this also fix https://github.com/iMicknl/ha-tahoma/issues/736?

tetienne:
> @tetienne will this also fix [iMicknl/ha-tahoma#736](https://github.com/iMicknl/ha-tahoma/issues/736)?

If he has been successfully logged at least once, it can fix it I guess. But if it’s from the boot, it won’t fix it.

tetienne:
Tested locally, backoff is not anymore involved when we refresh the token, meaning we are using the correct lister id. 
The only downside is we lost all the events when we switch the listener.

iMicknl (approved):
Ready to merge if tested. Will this be better than current situation? Or do we need to wait for other changes?

tetienne:
It will be better than the current situation indeed as we don't register multiple times the listener.

For the moment, the only downside as discussed, is the fact we can miss some events ever hour when we register the new listener.

I have an of fix, but I can do it in another pr.

pyoverkiz/client.py
```diff
@@ -519,6 +523,7 @@ async def fetch_events(self) -> list[Event]:
         Per-session rate-limit : 1 calls per 1 SECONDS period for this particular
         operation (polling)
         """
+        await self._refresh_token_if_expired()
```
iMicknl, line 526: Why do we call this for cozytouch login? We don't use it there?

pyoverkiz/client.py
tetienne: Because we want to refresh the `event_listener_id` before fetching the events. That’s the point of this PR.
If we don’t do that, we will call the `__post` with an expired token, and so an invalid `event_listener_id`. Even if we refresh the token into the `__post` method, the `event_listener_id` which is a parameter will still be wrong.

PR #349, opened 2022-01-30
Set last_update_time as optional for Location

tetienne:
See https://community.home-assistant.io/t/overkiz-api-and-somfy-api/61448/1715

### DIFF
--- a/pyoverkiz/models.py
+++ b/pyoverkiz/models.py
@@ -71,7 +71,7 @@ class Setup:
 @define(init=False, kw_only=True)
 class Location:
     creation_time: str
-    last_update_time: str
+    last_update_time: str | None = None
     city: str = field(repr=obfuscate_string, default=None)
     country: str = field(repr=obfuscate_string, default=None)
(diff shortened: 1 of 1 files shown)


pyoverkiz/models.py
```diff
@@ -71,7 +71,7 @@ def __init__(
 @define(init=False, kw_only=True)
 class Location:
     creation_time: str
-    last_update_time: str
+    last_update_time: str | None = None
```
iMicknl, line 74: Do we need to set it to None here as well? I think just below is enough.

tetienne: There are two different thing. Here we declare a property that’s an optional string, and below we declare a constructor taking as argument an optional string named `last_update_time`. They are only connected because in the constructor we assign the last one to the first one. 

By the way, you did the same for the `Setup` model :)

iMicknl: I thought for newer models I only set it to None in the init params, not in the class variables anymore.

However I am happy to merge if this works. Reviewing from my phone.

iMicknl (approved):

PR #379, opened 2022-02-20
Extend event enums

iMicknl:


tetienne (approved):
Where did you get all these new events?

iMicknl:
> Where did you get all these new events?

In a dump from the Somfy North America app I believe. There are way more, but I added some important ones.

PR #394, opened 2022-03-01
Add a lot of climate enums

iMicknl:


tetienne:
That's indeed a lot of new constants 😀

PR #411, opened 2022-03-14
Fix support for SIMU (LiveIn2) devices

daveixis:
Added support for SIMU (LiveIn2) devices - uses Somfy endpoint, but with old style authentication.


pyoverkiz/const.py
```diff
@@ -59,6 +59,12 @@
         manufacturer="Rexel",
         configuration_url="https://utilisateur.energeasyconnect.com/user/#/zone/equipements",
     ),
+    "simu_livein2": OverkizServer(  # uses https://ha101-1.overkiz.com
+        name="SIMU (LiveIn2)",
+        endpoint="https://tahomalink.com/enduser-mobile-web/enduserAPI/",
+        manufacturer="Somfy",
+        configuration_url="https://www.tahomalink.com",
```
iMicknl, line 66: Can you login via tahomalink.com?

Issue #321, opened 2022-01-17
Implement auth factory


iMicknl:
As discussed it would be good to implement some kind of authentication factory per server. 

Eventually it would be great to investigate

## Manufacturers
- [ ] **Atlantic Cozytouch** -> JWT / OAuth?

https://github.com/iMicknl/python-overkiz-api/blob/c5b8bf32e61ca81ac14a7e50379b24b5385107fa/pyoverkiz/client.py#L134-L137
Need to investigate if we can just pass this as a Bearer token to requests as well.

- [ ] **Hi Kumo** -> classic /login

- [ ] **Nexity Eugénie** -> ssoToken / OAuth?
https://github.com/iMicknl/python-overkiz-api/blob/c5b8bf32e61ca81ac14a7e50379b24b5385107fa/pyoverkiz/client.py#L140-L143
Need to investigate if we can just pass this as a Bearer token to requests as well.

- [ ] **Rexel Energeasy Connect** -> broken (uses Azure B2C without user/password flow)
Tracked in https://github.com/iMicknl/ha-tahoma/issues/591.

- [ ] **Somfy (Europe)** -> OAuth

Uses bearer token in header, need to handle token refresh
Would the Somfy official API client id / secret work as well?

- [ ] **Somfy (other)** -> classic /login

- [ ] **Local auth** (all, jailbroken)
Tracked in https://github.com/iMicknl/python-overkiz-api/issues/187

Endpoint is different (`/enduser-mobile-web/1/enduserAPI/`) and requires X-Auth-Token header.

```python3
self.headers.update ({
     "X-Auth-Token": self.api_token
})
```

tetienne:
I think we should expose an interface where a get and post methods must be implemented. And behind, the implementation does what it wants.

PR #441, opened 2022-04-04
Add local support for Somfy Developer Mode

iMicknl:
```python3
import asyncio
import time
import aiohttp

from pyoverkiz.client import OverkizClient
from pyoverkiz.const import SUPPORTED_SERVERS, OverkizServer

USERNAME = ""
PASSWORD = ""
LOCAL_GATEWAY = "gateway-1234-4567-8901.local"


async def main() -> None:

    token = ""

    async with OverkizClient(
        username=USERNAME, password=PASSWORD, server=SUPPORTED_SERVERS["somfy_europe"]
    ) as client:

        await client.login()
        gateways = await client.get_gateways()

        for gateway in gateways:
            # Generate tokens
            token = await client.generate_local_token(gateway.id)
            print(token)
            uuid = await client.activate_local_token(gateway_id=gateway.id, token=token, label="Home Assistant/local-dev")

    # # List all tokens
    # tokens = await client.get_local_tokens(gateway.id)

    # # Delete tokens
    # for token in tokens:
    #     print(token)
    #     print(await client.delete_local_token(gateway.id, token.uuid))

    # Local Connection
    session = aiohttp.ClientSession(connector=aiohttp.TCPConnector(ssl=False))

    async with OverkizClient(
        username="", password="", token=token, session=session, server=OverkizServer(
            name="Somfy TaHoma (local)",
            endpoint=f"https://{LOCAL_GATEWAY}:8443/enduser-mobile-web/1/enduserAPI/",
            manufacturer="Somfy",
            configuration_url=None,
        )
    ) as client:
        # print(await client.get_diagnostic_data())
        await client.login()

        print(await client.get_api_version())

        setup = await client.get_setup()
        print(setup)

        # devices = await client.get_devices()
        # print(devices)

        # for device in devices:
        #     print(f"{device.label} ({device.id}) - {device.controllable_name}")
        #     print(f"{device.widget} - {device.ui_class}")

        # while True:
        #     events = await client.fetch_events()
        #     print(events)

        #     time.sleep(2)

asyncio.run(main())
```


pyoverkiz/client.py
```diff
@@ -121,6 +127,18 @@ def __init__(
...
+            self._is_local = True
+            self._access_token = token
+
+            # To avoid security issue, add the following authority to
+            # your HTTPS client trust store: https://ca.overkiz.com/overkiz-root-ca-2048.crt
+            self._ssl_context = ssl.create_default_context(
+                cafile=os.path.dirname(os.path.realpath(__file__))
+                + "/overkiz-root-ca-2048.crt"
+            )
+            self._ssl_context = None
```
iMicknl, line 140: Blocked by https://github.com/Somfy-Developer/Somfy-TaHoma-Developer-Mode/issues/4

pyoverkiz/client.py
```diff
@@ -147,6 +165,16 @@ async def login(
         Authenticate and create an API session allowing access to the other operations.
         Caller must provide one of [userId+userPassword, userId+ssoToken, accessToken, jwt]
         """
+        # Local authentication
+        if self._is_local:
+            if register_event_listener:
+                await self.register_event_listener()
+            else:
+                # Call a simple endpoint to verify if our token is correct
+                await self.get_api_version()
```
iMicknl, line 174: This API endpoint doesn't have authentication 🤦‍♂️. I need to find another simple endpoint to verify a successful login.

pyoverkiz/client.py
iMicknl: SSL check removed for now. We can just pass a session that does not verify SSL.

pyoverkiz/client.py
```diff
@@ -122,6 +126,10 @@ def __init__(
 
         self.session = session if session else ClientSession()
 
+        if "/enduser-mobile-web/1/enduserAPI/" in server.endpoint:
```
tetienne, line 129: Would be better to add a property named for instance. `support_local`. WDYT?

iMicknl: I was in doubt about this one as well... In the end I would love to see something like this (https://github.com/iMicknl/python-overkiz-api/issues/321), where we define the auth mechanism in the `OverkizServer` definition. 

Eventually that would be even better, since we can than easily pick any authentication method, and we don't need an extra property just for local.

iMicknl: I simplified it in cfa2abc08e52eae553b25a4d369f0c720f8b1bdc, would that be ok for now? @tetienne

tetienne (approved):

tetienne (approved):

PR #453, opened 2022-04-11
Add local SSL CA check

iMicknl:


pyoverkiz/overkiz-root-ca-2048.crt
```diff
@@ -0,0 +1,20 @@
+-----BEGIN CERTIFICATE-----
```
tetienne, line 1: Where did you get it?

iMicknl: From https://ca.overkiz.com/overkiz-root-ca-2048.crt.

pyoverkiz/client.py
```diff
@@ -3,6 +3,8 @@
```
iMicknl, line 3: @tetienne lol, see CI. I will make the change to just ignore this, since we won't have time to refactor this huge module near term.

```
  ************* Module pyoverkiz.client
  pyoverkiz/client.py:1:0: C0302: Too many lines in module (1006/1000) (too-many-lines)`
```

iMicknl:
Just tested this PR and it works like a charm! Now we can do an SSL check on every request.

tetienne (approved):

pyoverkiz/client.py
tetienne: Well I have my pr where I did the refactor but we didn't test it 😅

PR #492, opened 2022-05-18
Add more constants related to Somfy Thermostat

tetienne:
See https://github.com/home-assistant/core/pull/67169/files


pyoverkiz/enums/state.py
```diff
@@ -124,12 +127,15 @@ class OverkizState(str, Enum):
...
     CORE_WIND_SPEED = "core:WindSpeedState"
+    FREEZE_MODE = "freezeMode"
+    FURTHER_NOTICE = "further_notice"
     HLRRWIFI_FAN_SPEED = "hlrrwifi:FanSpeedState"
     HLRRWIFI_LEAVE_HOME = "hlrrwifi:LeaveHomeState"
     HLRRWIFI_MAIN_OPERATION = "hlrrwifi:MainOperationState"
     HLRRWIFI_MODE_CHANGE = "hlrrwifi:ModeChangeState"
     HLRRWIFI_ROOM_TEMPERATURE = "hlrrwifi:RoomTemperatureState"
     HLRRWIFI_SWING = "hlrrwifi:SwingState"
+    INACTIVE = "inactive"
```
iMicknl, line 138: Is this a state or a state param?

pyoverkiz/enums/state.py
```diff
@@ -196,6 +203,7 @@ class OverkizState(str, Enum):
     RTDS_CONTROLLER_ORIGINATOR = "rtds:ControllerOriginatorState"
     RTDS_CONTROLLER_SENSING = "rtds:ControllerSensingState"
     RTDS_CONTROLLER_SIREN = "rtds:ControllerSirenState"
+    SLEEPING_MODE = "sleepingMode"
```
iMicknl, line 206: This is also not in the right place?

tetienne: Sorry, you are right. I misread your comment: https://github.com/iMicknl/python-overkiz-api/pull/492/files

iMicknl: You can use command.py OverkizCommandParam.

pyoverkiz/enums/state.py
tetienne: This is a state value, sorry. We don’t use enum for state value, right?

tetienne: I just checked, inactive and active are never used as command parameter, they are just a state value.

pyoverkiz/enums/state.py
tetienne: Done

pyoverkiz/enums/command.py
```diff
@@ -155,8 +157,10 @@ class OverkizCommandParam(str, Enum):
     EXTERNAL_GATEWAY = "externalGateway"
     FAN = "fan"
     FREE = "free"
+    FREEZE_MODE = "freezeMode"
     FROSTPROTECTION = "frostprotection"
     FULL = "full"
+    FURTHER_NOTICE = "further_notice"
```
iMicknl, line 163: Is this a Somfy inconsistency? Instead of `furtherNotice`.

iMicknl (approved):

pyoverkiz/enums/state.py
iMicknl: In theory we could still add it, so you can use the enum, however than the name OverkizCommandParam is a bit weird.

PR #582, opened 2022-08-27
Add method to execute scheduled scenario

PingouInfini:
Add new method and new associated event name

### DIFF
--- a/pyoverkiz/client.py
+++ b/pyoverkiz/client.py
@@ -313,7 +313,7 @@ class OverkizClient:
 
             jwt = jwt.strip('"')  # Remove surrounding quotes
 
-            return jwt
+            return cast(str, jwt)
 
     async def nexity_login(self) -> str:
         """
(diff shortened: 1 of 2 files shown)


tetienne (approved):
Thank you for your contribution!

iMicknl:
@PingouInfini could you run the pre-commits hooks / black? Or did you do this and is this an issue in the CI/CD here.

PingouInfini:
My bad. 
I was stuck with this mypy hook issue "pyoverkiz/client.py:316: error: Returning Any from function declared to return "str"" so I bypassed it.
New commit with fixes for hooks compliance
And I don't know why the CI/Python 8 is in error.

iMicknl:
```
  mypy.....................................................................Failed
  - hook id: mypy
  - exit code: 1
  
  pyoverkiz/client.py:316: error: Redundant cast to "str"  [redundant-cast]
```

I will have a look if I can fix it for you soon :-).

PingouInfini:
@iMicknl Everything seems fine with these latest commits.

PR #591, opened 2022-09-03
Add Ubiwizz server

iMicknl:
Add support for MyUbiwizz (https://play.google.com/store/apps/details?id=com.decelect.ubiwizz&hl=nl&gl=US)

### DIFF
--- a/pyoverkiz/const.py
+++ b/pyoverkiz/const.py
@@ -83,4 +83,10 @@ SUPPORTED_SERVERS: dict[str, OverkizServer] = {
         manufacturer="Somfy",
         configuration_url=None,
     ),
+    "ubiwizz": OverkizServer(
+        name="Ubiwizz",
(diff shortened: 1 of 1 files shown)


pyoverkiz/const.py
```diff
@@ -83,4 +83,10 @@
         manufacturer="Somfy",
         configuration_url=None,
     ),
+    "ubiwizz": OverkizServer(
```
tetienne, line 86: You found the url within the apk?

tetienne (approved):

pyoverkiz/const.py
iMicknl: Yes! I will verify this with the user that report it on HA community forum, before merging.

iMicknl: Confirmed working!

PR #615, opened 2022-09-24
backoff on ServerDisconnectedError also for single commands and scenario on 

floriangeigl:
same backoff for single commands and scenarios as for multi commands


tetienne (approved):
Good catch. Thx!

PR #639, opened 2022-10-17
Add gateway_id, device_address, subsystem_id and is_sub_device into device model

nyroDev:
As discussed, Adding new informations into Device object in order to be used elsewhere if needed


nyroDev:
Tests failed because json use "io://****-****-3293" instead of a well formed gatewayid like  "io://1234-5678-3293".  
Not sure if all JSON should be updated or something else should be done here.

tetienne (changes requested):
Hi, thx for your contribution.

Can you please add some tests?

pyoverkiz/models.py
```diff
@@ -181,7 +189,17 @@ def __init__(
...
+        match = re.search(DEVICE_REGEX, device_url)
+
+        self.protocol = Protocol(match.group('protocol'))
+        self.gateway_id = match.group('gatewayId')
+        self.device_address = match.group('deviceAddress')
+
+        if match.group('subsystemId'):
+            self.subsystem_id = int(match.group('subsystemId'))
+            if self.subsystem_id > 1:
+                self.is_sub_device = True
```
tetienne, line 202: What's the goal of such property?

tetienne:
> Tests failed because json use  ̀io:// **** - **** -3293 ̀ instead of a well formed gatewayid like  "io://1234-5678-3293".  
> 
> Not sure if all JSON should be updated or something else should be done here.

Indeed. You can update the JSON files with _real_ values.

pyoverkiz/models.py
nyroDev: There: https://github.com/home-assistant/core/blob/dev/homeassistant/components/overkiz/entity.py#L37

nyroDev:
> Hi, thx for your contribution.
> 
> Can you please add some tests?

@tetienne Done, I hope I did it correctly.

(sorry for all commits, I could not launch test on my machine)

pyoverkiz/models.py
```diff
@@ -181,7 +188,20 @@ def __init__(
         self.data_properties = data_properties
         self.type = ProductType(type)
         self.place_oid = place_oid
-        self.protocol = Protocol(self.device_url.split(":")[0])
+
+        match = re.search(DEVICE_REGEX, device_url)
+        assert match is not None
```
tetienne, line 193: Why use assert? We don’t need it.

tests/test_models.py
```diff
@@ -58,10 +58,78 @@
...
+    "uiClass": "RollerShutter",
+}
+
 STATE = "core:NameState"
 
 
 class TestDevice:
+    def test_base_url_parsing(self):
+        hump_device = humps.decamelize(RAW_DEVICES)
+        del hump_device["states"]
```
tetienne, line 115: Any reason to do this for your test?

pyoverkiz/models.py
```diff
@@ -181,7 +188,20 @@ def __init__(
...
+        self.protocol = Protocol(match.group("protocol"))
+        self.gateway_id = match.group("gatewayId")
+        self.device_address = match.group("deviceAddress")
+
+        self.subsystem_id = None
+        self.is_sub_device = False
+        if match.group("subsystemId"):
+            self.subsystem_id = int(match.group("subsystemId"))
+            if self.subsystem_id > 1:
+                self.is_sub_device = True
```
tetienne, line 204:
```suggestion
              self.is_sub_device = self.subsystem_id > 1
```

tetienne:
@nyroDev What’s blocking you to launch test on your machine? 
These should be enough:
```
poetry install
poetry shell
tox # or pytest
```

tests/test_models.py
nyroDev: Without it, mypay said:
`error: Item "None" of "Optional[Match[str]]" has no attribute "group"  [union-attr]`

nyroDev: It was done in an already existing test, I just copy/paste it.

tetienne: Don’t copy/paste blindly :)

You copy a test named `test_none_states` for which it makes sense to delete the... states.

pyoverkiz/models.py
```diff
@@ -24,6 +25,8 @@
 
 # pylint: disable=unused-argument, too-many-instance-attributes, too-many-locals
 
+DEVICE_REGEX = r"(?P<protocol>[a-zA-Z]+)://(?P<gatewayId>\d\d\d\d-\d\d\d\d-\d\d\d\d+)/(?P<deviceAddress>\w+)(#(?P<subsystemId>\d+))?"
```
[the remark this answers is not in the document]
nyroDev: Udated not exactly like your suggestion to make it work and handle the case of  / in and other elements.  
I added more tests for this kind of cases

pyoverkiz/models.py
```diff
@@ -24,6 +25,8 @@
 
 # pylint: disable=unused-argument, too-many-instance-attributes, too-many-locals
 
+DEVICE_URL_RE = r"(?P<protocol>.+)://(?P<gatewayId>\d\d\d\d-\d\d\d\d-\d\d\d\d)/(?P<deviceAddress>[^#]+)(#(?P<subsystemId>\d+))?"
```
tetienne, line 28:
```suggestion
DEVICE_URL_RE = r"(?P<protocol>.+)://(?P<gatewayId>\d{4}(-\d){3}/(?P<deviceAddress>[^#]+)(#(?P<subsystemId>\d+))?"
```

tetienne (approved):
Thx ! Nice improvement!

pyoverkiz/models.py
```diff
@@ -181,7 +188,19 @@ def __init__(
         self.data_properties = data_properties
         self.type = ProductType(type)
         self.place_oid = place_oid
-        self.protocol = Protocol(self.device_url.split(":")[0])
+
+        match = re.search(DEVICE_URL_RE, device_url)
```
iMicknl, line 192:
```suggestion
        # Split <protocol>://<gatewayId>/<deviceAddress>[#<subsystemId>] into multiple variables
        match = re.search(DEVICE_URL_RE, device_url)
```

pyoverkiz/models.py
```diff
@@ -181,7 +188,19 @@ def __init__(
...
+
+        self.subsystem_id = None
```
iMicknl, line 194: Do we need this? You already set it to `None` in the class declaration above?

pyoverkiz/models.py
```diff
@@ -181,7 +188,19 @@ def __init__(
...
+        self.is_sub_device = False
```
iMicknl, line 195: Same here

pyoverkiz/models.py
```diff
@@ -24,6 +25,8 @@
 
 # pylint: disable=unused-argument, too-many-instance-attributes, too-many-locals
 
+DEVICE_URL_RE = r"(?P<protocol>.+)://(?P<gatewayId>\d{4}-\d{4}-\d{4})/(?P<deviceAddress>[^#]+)(#(?P<subsystemId>\d+))?"
```
iMicknl, line 28:
```suggestion
# <protocol>://<gatewayId>/<deviceAddress>[#<subsystemId>]
DEVICE_URL_RE = r"(?P<protocol>.+)://(?P<gatewayId>\d{4}-\d{4}-\d{4})/(?P<deviceAddress>[^#]+)(#(?P<subsystemId>\d+))?"
```

pyoverkiz/models.py
```diff
@@ -181,7 +188,19 @@ def __init__(
...
-        self.protocol = Protocol(self.device_url.split(":")[0])
+
+        match = re.search(DEVICE_URL_RE, device_url)
+
+        self.subsystem_id = None
+        self.is_sub_device = False
+        if match:
+            self.protocol = Protocol(match.group("protocol"))
+            self.gateway_id = match.group("gatewayId")
+            self.device_address = match.group("deviceAddress")
```
iMicknl, line 199: We check for `if match`, but these variables are not optional in the device object. How do we handle this when we don't have a match? (can this happen?). Do we always have matches in all our test files?

iMicknl (commented):
Great addition! Some minor remarks, but looks good to me. Good to have tests, since this feature can break the integration for users if we fail on the matches.

pyoverkiz/models.py
nyroDev: I think that's the case.  
In test_client, a Setup is created with all json files.  
On init of Setup, we have   
`self.devices = [Device(**d) for d in devices]`
So I guess yes, all elements from test are correct.

I also reviewed all "deviceURL" in tests file adn all are formed corresponding to the regex.

pyoverkiz/models.py
nyroDev: With it, test don't pass:
" AttributeError: 'Device' object has no attribute 'subsystem_id'"

pyoverkiz/models.py
iMicknl: Ok, but this `if match` should not be necessary in that case? If there is no match, self.gateway_id and self.device_address won't be set.

nyroDev: See here, because of mypy...
https://github.com/iMicknl/python-overkiz-api/pull/639#pullrequestreview-1145241149

iMicknl: I don't see what the relation is with this `if match`?

nyroDev: Without it, mypy tells me:  
`Item "None" of "Optional[Match[str]]" has no attribute "group"`

pyoverkiz/models.py
tetienne: > Do we need this? You already set it to `None` in the class declaration above?

As we override the constructor, we bypass the attr.ib declarations.

pyoverkiz/models.py
tetienne: What's surprise me is that mypy does not complain about missing gateway id if there is no match...

nyroDev: I guess it's because of :  
https://docs.python.org/3/library/re.html#re.Match.group

> If the regular expression uses the (?P<name>...) syntax, the groupN arguments may also be strings identifying groups by their group name. If a string argument is not used as a group name in the pattern, an [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError) exception is raised.

No match found, exception is raised.

pyoverkiz/models.py
nyroDev: What about the `@define(init=False)` existing on the class definition?  
Could it be related?

iMicknl:
@nyroDev I tried to push to this PR, but it seems that we cannot do that. See commit below for my changes, to at least make sure we test all the JSON files for a valid gateway_id.

https://github.com/nyroDev/python-overkiz-api/compare/add-device_address-and-subsystem_id...iMicknl:python-overkiz-api:add-device_address-and-subsystem_id

nyroDev:
> @nyroDev I tried to push to this PR, but it seems that we cannot do that. See commit below for my changes, to at least make sure we test all the JSON files for a valid gateway_id.
> 
> [nyroDev/python-overkiz-api@add-device_address-and-subsystem_id...iMicknl:python-overkiz-api:add-device_address-and-subsystem_id](https://github.com/nyroDev/python-overkiz-api/compare/add-device_address-and-subsystem_id...iMicknl:python-overkiz-api:add-device_address-and-subsystem_id)

I updated the code like you suggested, all tests passed.

nyroDev:
> > @nyroDev I tried to push to this PR, but it seems that we cannot do that. See commit below for my changes, to at least make sure we test all the JSON files for a valid gateway_id.
> > [nyroDev/python-overkiz-api@add-device_address-and-subsystem_id...iMicknl:python-overkiz-api:add-device_address-and-subsystem_id](https://github.com/nyroDev/python-overkiz-api/compare/add-device_address-and-subsystem_id...iMicknl:python-overkiz-api:add-device_address-and-subsystem_id)
> 
> I updated the code like you suggested, all tests passed.

I just realized that I didn't mention @iMicknl on my last comment.
So here it is ;)

iMicknl (approved):
Let's make sure we bump the version number to a minor update.

PR #661, opened 2022-11-07
Add base_device_url and corresponding tests on Device model

nyroDev:
After starting the implementation on HA component, I figured out that we where missing the base_device_url.  
We could recreate it there, but I think it's better to have always available on the device info.

### DIFF
--- a/pyoverkiz/models.py
+++ b/pyoverkiz/models.py
@@ -144,6 +144,7 @@ class Device:
     enabled: bool
     label: str = field(repr=obfuscate_string)
     device_url: str = field(repr=obfuscate_id)
+    base_device_url: str = field(repr=obfuscate_id)
     gateway_id: str = field(repr=obfuscate_id)
     device_address: str = field(repr=obfuscate_id)
     subsystem_id: int | None = None
@@ -200,6 +201,10 @@ class Device:
             self.protocol = Protocol(match.group("protocol"))
             self.gateway_id = match.group("gatewayId")
             self.device_address = match.group("deviceAddress")
+
+            self.base_device_url = (
+                self.protocol + "://" + self.gateway_id + "/" + self.device_address
+            )
 
             if match.group("subsystemId"):
                 self.subsystem_id = int(match.group("subsystemId"))
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -151,6 +151,7 @@ class TestOverkizClient:
                 assert device.gateway_id
                 assert device.device_address
                 assert device.protocol
+                assert device.base_device_url
 
 
 class MockResponse:
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -65,9 +65,10 @@ STATE = "core:NameState"
 
 class TestDevice:
     @pytest.mark.parametrize(
-        "device_url, protocol, gateway_id, device_address, subsystem_id, is_sub_device",
+        "device_url, base_device_url, protocol, gateway_id, device_address, subsystem_id, is_sub_device",
         [
             (
+                "io://1234-5678-9012/10077486",
                 "io://1234-5678-9012/10077486",
                 Protocol.IO,
                 "1234-5678-9012",
@@ -77,6 +78,7 @@ class TestDevice:
             ),
             (
                 "io://1234-5678-9012/10077486#8",
+                "io://1234-5678-9012/10077486",
                 Protocol.IO,
                 "1234-5678-9012",
                 "10077486",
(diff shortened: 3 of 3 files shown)


nyroDev:
@iMicknl 
Once this PR is merged and published, I'll be able to add new PR on HA.
Everything is [ready here](https://github.com/nyroDev/home-assistant-core/commit/7d1a4328461064c7acfa7b3335d4ca4318431796), I'll juste have to update the version number

iMicknl (approved):
Let's wait for a 2nd approver, and then we will merge + release.

tetienne:
I don't see the pro of such property. It will be confusing too have a device url and a base url. From the name, the diff is not obvious.

nyroDev:
> I don't see the pro of such property. It will be confusing too have a device url and a base url. From the name, the diff is not obvious.

@tetienne This is the name [currently used in HomeAssistant](https://github.com/home-assistant/core/blob/dev/homeassistant/components/overkiz/entity.py#L31) (as well as [in HA-Tahoma](https://github.com/iMicknl/ha-tahoma/blob/master/custom_components/tahoma/entity.py#L30))

nyroDev:
Hi @tetienne did you have time to read my replies?
Any suggestion for the name maybe?

nyroDev:
Hi @tetienne did you have time to read my replies?
Any suggestion for the name maybe?

tetienne:
I have a suggestion related to the various addition we did on the Device model. 
As we have IMO too many properties directly within Device, what do you think if we create a DeviceUrl class:

```python
class DeviceUrl:

  def __init__(device_url: str):
    self.device_url = device_url
    self.base = ...
    self.index = ...
```

It will move the boilerplate, and ease the logic. Of course, we should declare it as a breaking change.

nyroDev:
> I have a suggestion related to the various addition we did on the Device model. As we have IMO too many properties directly within Device, what do you think if we create a DeviceUrl class:
> 
> ```python
> class DeviceUrl:
> 
>   def __init__(device_url: str):
>     self.device_url = device_url
>     self.base = ...
>     self.index = ...
> ```
> 
> It will move the boilerplate, and ease the logic. Of course, we should declare it as a breaking change.
> 
> In the client code:
> 
> ```python
> device.url.full
> device.url.base
> device.url.index
> ```

@tetienne makes sense to me, but will it be worth all the rewriting?

Anyway, if accepted, we should have on this DeviceUrl : 
- full
- base (same as full without `#index`) 
- protocol
- gateway
- address
- index (int)
- is_sub (bool)

@iMicknl what do you think?

nyroDev:
@iMicknl @tetienne What should I do here?

tetienne:
IMO there is too many properties now within Device and we don’t match anymore what’s returned by Overkiz.

While previously I suggested to change the type of device_url, it will introduce a breaking change. To avoid this, we can create a DeviceUrlParser class:


```python
class DeviceUrlParser:
  def __init__(device_url: str):
    self.device_url = device_url
    self.base = ...
    self.index = ...
    self.protocol = ...
    self.gateway = ...
    ...
```

So in the client code, you retrieve the device_url, and then instantiate a DeviceUrlParser. We keep the model close from Overkiz, and we introduce an utility class.
In the changelog, we can flag as deprecated the replaced properties within Device.

nyroDev:
> IMO there is too many properties now within Device and we don’t match anymore what’s returned by Overkiz.
> 
> While previously I suggested to change the type of device_url, it will introduce a breaking change. To avoid this, we can create a DeviceUrlParser class:
> 
> ```python
> class DeviceUrlParser:
>   def __init__(device_url: str):
>     self.device_url = device_url
>     self.base = ...
>     self.index = ...
>     self.protocol = ...
>     self.gateway = ...
>     ...
> ```
> 
> So in the client code, you retrieve the device_url, and then instantiate a DeviceUrlParser. We keep the model close from Overkiz, and we introduce an utility class. In the changelog, we can flag as deprecated the replaced properties within Device.

It will still be a breaking change, as you're removing properties that was there before.

nyroDev:
@iMicknl What do you think about this PR and the related on HA [ready](https://github.com/home-assistant/core/compare/dev...nyroDev:home-assistant-core:overkiz/pyoverkiz_update_device_info)

iMicknl:
@nyroDev sorry for the long due merge/review. Can you perhaps rebase on main and see if CI/CD works again? I will have a look at this PR again. If it doesn't have any breaking changes, I am fine with it.

nyroDev:
> @nyroDev sorry for the long due merge/review. Can you perhaps rebase on main and see if CI/CD works again? I will have a look at this PR again. If it doesn't have any breaking changes, I am fine with it.

Done.

Issue #681, opened 2022-11-26
API URL and Client ID has changed for Atlantic Cozy Touch API


adetante:
The old API Manager URL `api.groupe-atlantic.com` is not working anymore.

PR #682, opened 2022-11-26
New API Manager URL & Client ID for Atlantic

adetante:
Fix issue #681

### DIFF
--- a/pyoverkiz/client.py
+++ b/pyoverkiz/client.py
@@ -304,7 +304,7 @@ class OverkizClient:
 
         # Request JWT
         async with self.session.get(
-            COZYTOUCH_ATLANTIC_API + "/gacoma/gacomawcfservice/accounts/jwt",
+            COZYTOUCH_ATLANTIC_API + "/magellan/accounts/jwt",
             headers={"Authorization": f"Bearer {token['access_token']}"},
         ) as response:
             jwt = await response.text()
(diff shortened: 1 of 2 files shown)


iMicknl:
@adetante can you elaborate on where you found the new credentials / API? Did you take this from the Android app?

We need to double check if this does not break other accounts, but I hope to merge it tomorrow, if I am able to test.

adetante:
Hi @iMicknl
Yes, from the latest version of the Android app (3.6.0) I got the following values from `BuildConfig`:

```
public static final PROD_END_POINT = "/magellan"
public static final PROD_END_POINT_OLD = "/gacoma/gacomawcfservice"
public static final PROD_URL = "https://apis.groupe-atlantic.com"
public static final PROD_URL_OLD = "https://api.groupe-atlantic.com"
```

I use this python lib to get cozy touch data in my home assistant dashboard, and the API calls do not work anymore with the old URL.

For the credentials, unfortunately I can't inspect the calls made by the Android app at the moment.
I got the value from this PR: https://github.com/geogeocutcut/jeedom-cozytouch-plugin/pull/14/files

And indeed in my local installation these new credentials work.
But I can't guarantee that it's the one currently used by the Android app...

iMicknl:
Thanks @adetante! I will have a look :).

iMicknl:
@adetante I was able to verify this with 2 users and it seems to work.

Thanks for your contribution!

iMicknl (approved):

PR #683, opened 2022-11-27
Fix PR-labeler action for forks

iMicknl:


.github/workflows/pr-labeler.yml
```diff
@@ -1,6 +1,6 @@
 name: PR Labeler
 on:
-  pull_request:
+  pull_request_target:
```
tetienne, line 3: We don't need both?

tetienne (approved):

.github/workflows/pr-labeler.yml
iMicknl: Nope, apparently, `pull_request_target` will work for both.

iMicknl: However, it seems that it is not triggering on this PR anymore.

PR #686, opened 2022-11-29
Add private person domain to CozyTouch login

iMicknl:


iMicknl:
Tested with 2 different user logins. Got this from the Jeedom plugin.

PR #702, opened 2022-12-10
Fix device url regex for more gateway-id

nyroDev:
This is a fix for https://github.com/home-assistant/core/issues/83678
It simplify how the regex is done to be less restrictive on the gateway-id


tests/test_models.py
```diff
@@ -131,6 +131,22 @@ class TestDevice:
...
+                None,
+                False,
+            ),
+            (
+                "eliot://ELIOT-000000000000000000000000000ABCDE/00000000000000000000000000125abc#1",
+                Protocol.ELIOT,
+                "ELIOT-000000000000000000000000000ABCDE",
+                "00000000000000000000000000125abc",
+                1,
+                False,
```
tetienne, line 148: Can you please also add a test where the device url does not match?

tetienne (changes requested):

tests/test_models.py
tetienne: The idea is too avoid bad surprise if we retrieve a new kind of device url.

> I don't understand how I should do it
Just add a new set of parameters to your test

> how should I define which result to test against for a bad result
That’s you that defined the result when you implemented the logic in Device class :)
If any error is raised, it means we are doing something bad.

I guess, the following set of parameters should work:

```
(
                "foo://whatever-blah/12",
                None,
                "whatever-blah,
                "12",
                None,
                False,
            ),
```

```
(
                "foo://whatever",
                None,
                None,
                None,
                None,
                False,
            ),
```

nyroDev: Done.
I have to update init of vars to comply with mypy and wrong test case

tests/test_models.py
```diff
@@ -131,6 +131,39 @@ class TestDevice:
...
+                "foo://whatever-blah/12",
+                Protocol.UNKNOWN,
+                "whatever-blah",
+                "12",
+                None,
+                False,
+            ),
+            (
+                "foo://whatever",
+                None,
```
tetienne, line 161: Later we can do a fix to at least return the protocol

tetienne (approved):
Perfect!

PR #709, opened 2022-12-14
Add SET_SCHEDULING_TYPE command

nyroDev:
This will be used for https://github.com/home-assistant/core/pull/84010

The more I'm using pyoverkiz, the more I'm wondering: Why do we need to maintain a list of all commands here?  
This is pretty heavy to have to :  
1. MR here
2. Wait for release
3. Do the requirements update on HA

Essentially when, like this MR, it concerns only 1 command...


tetienne (approved):

tetienne:
Adding command here allow sharing with the others, and we enforce typing

PR #767, opened 2023-02-14
Split login logic

### LINKED ISSUE #321: Implement auth factory

tetienne:
Fix #321

The new main entrypoint is now `Overkiz.get_client_for`. 

`OverkizClient` does not manage anymore the session, it’s up to the calling code to manage it. See my example below. 

A StrEnum `Server` has been created to list all the existing supported servers.

```python
import asyncio
import time

from aiohttp import ClientSession

from pyoverkiz.clients.overkiz import OverkizClient
from pyoverkiz.const import Server
from pyoverkiz.overkiz import Overkiz

USERNAME = ""
PASSWORD = ""


async def main() -> None:

    async with ClientSession() as session:
        client = Overkiz.get_client_for(
            Server.SOMFY_EUROPE, USERNAME, PASSWORD, session
        )
        try:
            await client.login()
        except Exception as exception:  # pylint: disable=broad-except
            print(exception)
            return

        gateways = await client.get_gateways()
        token = await client.generate_local_token(gateways[0].id)
        print(token)
        await client.activate_local_token(gateways[0].id, token, "pyoverkiz")

        local_client: OverkizClient = Overkiz.get_client_for(
            Server.SOMFY_DEV_MODE, gateways[0].id, token, session
        )

        devices = await local_client.get_devices()

        for device in devices:
            print(f"{device.label} ({device.id}) - {device.controllable_name}")
            print(f"{device.widget} - {device.ui_class}")

        await local_client.register_event_listener()

        while True:
            events = await local_client.fetch_events()
            print(events)

            time.sleep(2)


asyncio.run(main())
```


pyoverkiz/servers/overkiz_server.py
```diff
@@ -0,0 +1,166 @@
...
+    name: str
+    endpoint: str
+    manufacturer: str
+    session: ClientSession
+
+    configuration_url: str | None
+
+    @abstractmethod
+    async def login(self, username: str, password: str) -> bool:
+        """Login to the server."""
```
iMicknl, line 45:
```suggestion
        """Login to the server."""
        pass
```
Isn't this the writing style for 'interfaces' in Python?

iMicknl (commented):
Looks great already! Would be good to think how this could work for LocalServers as well.

pyoverkiz/servers/overkiz_server.py
tetienne: Body can be empty, the doctring is enough for readability.

pyoverkiz/clients/somfy_local.py
```diff
@@ -0,0 +1,43 @@
...
+from typing import Any
+
+from pyoverkiz.clients.overkiz import OverkizClient
+from pyoverkiz.types import JSON
+
+
+class SomfyLocalClient(OverkizClient):
+    async def _login(self, username: str, password: str) -> bool:
+        """There is no login for Somfy Local"""
+        return True
```
iMicknl, line 12: Should we not make an alternative call here? Can we just return True while we aren't sure if the token is correct?

pyoverkiz/clients/somfy_local.py
```diff
@@ -0,0 +1,43 @@
...
+        """Make a GET request to the OverKiz API"""
+
+        return await super().get(path, ssl=False)
+
+    async def post(
+        self,
+        path: str,
+        payload: JSON | None = None,
+        data: JSON | None = None,
+        _ssl: bool = False,
```
iMicknl, line 32: Should we set the SSL=False not in the session? Isn't this something we should not define hardcoded in our get() implementation.

pyoverkiz/const.py
```diff
@@ -1,104 +1,150 @@
...
         endpoint="https://ha101-1.overkiz.com/enduser-mobile-web/enduserAPI/",
         manufacturer="Somfy",
         configuration_url="https://www.tahomalink.com",
+        session=session,
+        username=username,
+        password=password,
+    ),
+    "somfy_europe_local": lambda gateway_id, token, session: SomfyLocalClient(
+        name="Somfy Local(Europe)",
+        endpoint=f"https://gateway-{gateway_id}.local:8443/enduser-mobile-web/1/enduserAPI/",
```
iMicknl, line 116: This is not flexible enough I would say. There will be users (like me) where bonjour is not working (e.g. they have a VPN / different VNET). It would be good if we can specify a full URL, thus an IP or a .local address is both possible.

pyoverkiz/const.py
```diff
@@ -1,104 +1,150 @@
...
         name="Somfy (Europe)",
         endpoint="https://ha101-1.overkiz.com/enduser-mobile-web/enduserAPI/",
         manufacturer="Somfy",
         configuration_url="https://www.tahomalink.com",
+        session=session,
+        username=username,
+        password=password,
+    ),
+    "somfy_europe_local": lambda gateway_id, token, session: SomfyLocalClient(
+        name="Somfy Local(Europe)",
```
iMicknl, line 115:
```suggestion
        name="Somfy Developer Mode (Local API)",
```

pyoverkiz/clients/somfy_local.py
tetienne: Well, this login method should not be called when you used the Local API. I can instead raise an exception too make this more clear.

pyoverkiz/clients/somfy_local.py
tetienne: Yes indeed, I forgot to push my latest commits.

pyoverkiz/const.py
tetienne: Hum, that’s good to now. So on HA side, you have to enter manually the IP of your box?

pyoverkiz/const.py
tetienne: Done.

pyoverkiz/const.py
iMicknl: Yes

iMicknl (commented):
Had a quick look! Thanks, great improvement.

pyoverkiz/overkiz.py
```diff
@@ -0,0 +1,161 @@
...
+        username=username,
+        password=password,
+    ),
+    Server.SOMFY_DEV_MODE: lambda domain, token, session: SomfyLocalClient(
+        name="Somfy Developer Mode (Local API)",
+        endpoint=f"https://{domain}:8443/enduser-mobile-web/1/enduserAPI/",
+        manufacturer="Somfy",
+        configuration_url=None,
+        session=session,
+        username=domain,  # not used
```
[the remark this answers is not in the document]
tetienne: You can check my lastest modifications, username and password are no more within the parent class, but are defined within the specific implementation.

pyoverkiz/overkiz.py
```diff
@@ -0,0 +1,161 @@
...
+        session=session,
+        username=username,
+        password=password,
+    ),
+}
+
+
+class Overkiz:
+    @staticmethod
+    def get_client_for(
```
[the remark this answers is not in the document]
tetienne: Done.

pyoverkiz/clients/somfy_local.py
```diff
@@ -0,0 +1,20 @@
...
+from pyoverkiz.clients.overkiz import OverkizClient
+
+
+@define(kw_only=True)
+class SomfyLocalClient(OverkizClient):
+
+    _ssl: bool = field(default=False, init=False)
+    token: str = field(repr=lambda _: "***")
+
+    async def _login(self) -> bool:
```
iMicknl, line 14: How can we verify if the local login works?

tetienne: We can call any command to see if it works. 
But I don’t think we should put a check here. There is no login endpoint for the local api, and so we should not call it.

PR #841, opened 2023-05-02
Add failure type 197 (CMDDEPRECATED) to enums

iMicknl:
Fixes https://github.com/home-assistant/core/issues/91206


pyoverkiz/enums/general.py
```diff
@@ -126,6 +126,7 @@ class FailureType(IntEnum):
     PRIORITY_LOCK__NO_SECURITY_DEVICE = 192
     PRIORITY_LOCK__DEAD_SENSOR = 193
     PRIORITY_LOCK__UNKNOWN_ERROR = 194
+    CMDDEPRECATED = 197
```
tetienne, line 129: Oh! Nice to see they start to depreciate some commands. In which case did you see it?

iMicknl: https://github.com/home-assistant/core/issues/91206, but not more context than this :(

tetienne: Hope we will get an answer.

PR #843, opened 2023-05-03
Fix detection of error messages.

shypike:
Don't try to do exact matches, but use "in".
One issue being that the message usually looks like:
  '"No registered event listener."'
(Note the extra quotes.)

### DIFF
--- a/pyoverkiz/client.py
+++ b/pyoverkiz/client.py
@@ -809,7 +809,7 @@ class OverkizClient:
             ) from error
 
         if result.get("errorCode"):
-            message = result.get("error")
+            message = result.get("error").strip("'")
 
             # {"errorCode": "AUTHENTICATION_ERROR",
             # "error": "Too many requests, try again later : login with xxx@xxx.tld"}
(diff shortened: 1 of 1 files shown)


shypike:
This will make recovery of the "No registered event listener" error during local access of a Somfy box actually work.
The recovery code is correct, but the "==" matching spoils it here.

iMicknl:
Thanks @shypike, we will have a look! Interesting that they throw a slight different error locally... 

How are you using the local integration? Via your own Python code or via HA?

iMicknl:
@shypike we actually have a version of Overkiz with a local integration ready, for months, but it is not polished. If you are want to contribute, you are more than welcome. https://github.com/home-assistant/core/pull/71644

See https://community.home-assistant.io/t/overkiz-integration-local-api-development-testers-topic/485264

iMicknl:
@vlebourl, @egguy do we want to merge all these changes in, or just the one that is required for this usecase?

shypike:
@iMicknl I'll have a look, but with summer coming :)

tetienne (approved):

iMicknl (changes requested):
@shypike we would prefer to only have the messages changes where there is no exact match, for clarity. Or even add exact string matches in the if statement, instead of an `in message`.

Would you be able to make this change, so we can merge it in?

shypike:
@iMicknl What is not clear to me is whether the "No registered event listener" is unique for local access. If it's valid for both local and remote, using exact string matches will not work. Unless your original code was already incorrect for remote.
Personally I'm no fan of exact strings matches because it makes it more likely to fail due to trivial changes.
I would go for an "in" for the "No registered event listener" message and no change to other strings.
BTW: my local hack (using the above changes) is rock solid has been running successfully for three weeks now. Unlike the remote version which is always hampered by the sub-par Somfy servers.

iMicknl:
I will try to simulate the error on cloud API to see if this is incorrect in the client library, or if this is something that is using another error message in the local API.

The reason we would like exact matches is too be as strict as possible, if Overkiz/Somfy would change the error messages, we probably would need changes anyway, thus it is good to have it fail. We rather have everything as strict as possible.

Great to hear that your local hack is working great! We still need to full implement this in HA. Can you perhaps verify if you have the same issue? https://github.com/Somfy-Developer/Somfy-TaHoma-Developer-Mode/issues/102

shypike:
@iMicknl I have mostly RTS devices and only two IO screens. The RTS devices don't give any feedback, so always have state unknown. I'll keep an eye on IO. However, I've tried to set debug mode on but nothing is logged.
```
logger:
  default: error
  logs:
    pyoverkiz: debug
    homeassistant.components.overkiz: debug
```
The above lines in the config should do the trick. What am I doing wrong?

shypike:
My last commit just strips off outer single quotes. That should care of the difference between local and remote messages.

iMicknl (approved):

shypike:
Pfff, I stripped the wrong kind of quotes. When I read my first text again, I see that "local" adds double quotes and not single quotes.
So the best solution would be:
```
.strip("\"'")
```

iMicknl:
@shypike can you save some JSON files of the error? So we can add testcases for this.

PR #855, opened 2023-05-19
Fix diagnostics issue and add diagnostics test

iMicknl:


pyoverkiz/obfuscate.py
```diff
@@ -26,6 +26,8 @@ def obfuscate_sensitive_data(data: dict[str, Any]) -> JSON:
     """Mask Overkiz JSON data to remove sensitive data"""
     mask_next_value = False
 
+    print(data)
```
tetienne, line 29: Did you forget to remove it?

tetienne (commented):
Thx 😀

pyoverkiz/obfuscate.py
iMicknl: Thanks, fixed!

tetienne (approved):

PR #865, opened 2023-05-28
Fix local error messages

iMicknl:
Follow up on https://github.com/iMicknl/python-overkiz-api/pull/843#issuecomment-1556157440.

@shypike can you verify if this is the right fix?

### DIFF
--- a/pyoverkiz/client.py
+++ b/pyoverkiz/client.py
@@ -809,7 +809,7 @@ class OverkizClient:
             ) from error
 
         if result.get("errorCode"):
-            message = result.get("error").strip("'")
+            message = result.get("error").strip('"')
 
             # {"errorCode": "AUTHENTICATION_ERROR",
             # "error": "Too many requests, try again later : login with xxx@xxx.tld"}


shypike:
It should be OK, but I'll add it to my own installation. And I will try to catch the error condition.
I can give you feedback tomorrow.

iMicknl:
> It should be OK, but I'll add it to my own installation. And I will try to catch the error condition. I can give you feedback tomorrow.

Thanks! Let me know :) would be good to catch some JSON files so we can eventually add tests.

pyoverkiz/client.py
```diff
@@ -809,7 +809,7 @@ async def check_response(response: ClientResponse) -> None:
             ) from error
 
         if result.get("errorCode"):
-            message = result.get("error").strip("'")
+            message = result.get("error").strip('"')
```
tetienne, line 812: If we don’t add any test, it would be nice to add as comment an example of why we trip the ".

iMicknl: Agree! Preferably both 😄.

shypike:
It looks like this one is the only deviation:

```
{"error":"\"No registered event listener.\"","errorCode":"UNSPECIFIED_ERROR"}
```

shypike:
I've also tested the same issue with the cloud connection.
Conclusion.
Cloud says:
```
{"errorCode":"UNSPECIFIED_ERROR","error":"No registered event listener"}
```
Local says:
```
{"error":"\"No registered event listener.\"","errorCode":"UNSPECIFIED_ERROR"}
```

iMicknl:
Thanks @shypike. Do you know which HTTP code these requests have?

shypike:
I've added the error codes to the original message,
Of course the required change is not just stripping with .strip("\"'.") but also removing the trailing "." from the string comparisons.

iMicknl:
Thanks! @shypike, I implemented it in #870.

PR #888, opened 2023-06-25
Add APIType, enums for server names and util to generate (local) OverkizServer class

iMicknl:
Adds an enum + variable to the client class to make it easier to detect if local or cloud API is used.

### DIFF
--- a/pyoverkiz/client.py
+++ b/pyoverkiz/client.py
@@ -132,7 +132,7 @@ class OverkizClient:
 
         self.session = session if session else ClientSession()
 
-        if "/enduser-mobile-web/1/enduserAPI/" in self.server.endpoint:
+        if LOCAL_API_PATH in self.server.endpoint:
             self.api_type = APIType.LOCAL
         else:
             self.api_type = APIType.CLOUD
@@ -164,11 +164,12 @@ class OverkizClient:
         Caller must provide one of [userId+userPassword, userId+ssoToken, accessToken, jwt]
         """
         # Local authentication
-        if LOCAL_API_PATH in self.server.endpoint:
+        if self.api_type == APIType.LOCAL:
             if register_event_listener:
                 await self.register_event_listener()
             else:
                 # Call a simple endpoint to verify if our token is correct
+                # Since local API does not have a /login endpoint
                 await self.get_gateways()
 
             return True
--- a/pyoverkiz/client.py
+++ b/pyoverkiz/client.py
@@ -16,6 +16,7 @@ from aiohttp import ClientResponse, ClientSession, FormData, ServerDisconnectedE
 from botocore.config import Config
 from warrant_lite import WarrantLite
 
+from pyoverkiz.enums import APIType
 from pyoverkiz.const import (
     COZYTOUCH_ATLANTIC_API,
     COZYTOUCH_CLIENT_ID,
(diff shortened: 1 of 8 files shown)


pyoverkiz/client.py
```diff
@@ -129,6 +131,12 @@ def __init__(
 
         self.session = session if session else ClientSession()
 
+        self.api_type = (
```
tetienne, line 134: I'm not a big fan of this style.

iMicknl: What would you prefer? Just a regular if/else statement?

iMicknl:
@tetienne I am thinking about porting this as well: https://github.com/home-assistant/core/blob/50ae1d8e2735ade85cb68e08d9022fc6ac5bf899/homeassistant/components/overkiz/config_flow.py#L39-L55, to have more code in our client than in the integration.

pyoverkiz/enums/server.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@unique
+class APIType(str, Enum):
+    CLOUD = "cloud"
+    LOCAL = "local"
+
+
+@unique
+class Server(str, Enum):
```
iMicknl, line 11: @tetienne I already added this one from your PR. Is a great add and would be good to have this in soon, before we fully migrate to the "v2" style of this client.

pyoverkiz/client.py
```diff
@@ -129,6 +131,11 @@ def __init__(
 
         self.session = session if session else ClientSession()
 
+        if "/enduser-mobile-web/1/enduserAPI/" in self.server.endpoint:
```
tetienne, line 134: Can you create a constant?

pyoverkiz/const.py
```diff
@@ -16,86 +17,92 @@
 SOMFY_CLIENT_ID = "0d8e920c-1478-11e7-a377-02dd59bd3041_1ewvaqmclfogo4kcsoo0c8k4kso884owg08sg8c40sk4go4ksg"
 SOMFY_CLIENT_SECRET = "12k73w1n540g8o4cokg0cw84cog840k84cwggscwg884004kgk"
 
+SERVERS_THAT_SUPPORT_LOCAL_API = [
```
tetienne, line 20: What about?

```suggestion
SERVERS_W_LOCAL_API = [
```

pyoverkiz/enums/server.py
tetienne: Perfect, I will see if I can backport more changes without breaking anything

pyoverkiz/const.py
iMicknl: I am not a huge fan of abbreviations like `W`, what about `SERVERS_WITH_LOCAL_API`? Or `DEV_MODE_SERVERS`?

pyoverkiz/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    """Generate OverkizServer class for connection with a local API (Somfy Developer mode)."""
+    return OverkizServer(
+        name=name,
+        endpoint=f"https://{host}{LOCAL_API_PATH}",
+        manufacturer=manufacturer,
+        configuration_url=configuration_url,
+    )
+
+
+def is_overkiz_gateway(gateway_id: str) -> bool:
```
tetienne, line 24: What’s an additional gateway?

tetienne (approved):

pyoverkiz/utils.py
iMicknl: Some users have a thermostat + protect system, which is also listed as a gateway. We have a JSON definition of one of them in our fixtures folder.

PR #897, opened 2023-06-30
Add setup option endpoint

iMicknl:
First implementation of https://github.com/Somfy-Developer/Somfy-TaHoma-Developer-Mode/issues/90#issuecomment-1614415702

- [x] Add tests
- [ ] get_setup_option_parameter(option, parameter) will return ` {'errorCode': 'UNSPECIFIED_ERROR', 'error': 'Internal Error'}` when option does not exist. If parameter does not exist, it will return None. 
- [ ] Add helper for local setup that will return a bool?


pyoverkiz/client.py
```diff
@@ -758,6 +760,61 @@ async def execute_scheduled_scenario(self, oid: str, timestamp: int) -> str:
...
+
+        return None
+
+    @backoff.on_exception(
+        backoff.expo,
+        (NotAuthenticatedException, ServerDisconnectedError),
+        max_tries=2,
+        on_backoff=relogin,
+    )
+    async def get_setup_option_parameter(
```
tetienne, line 804: Nice!

iMicknl: Except for the fact that this throws a very generic and ugly exception on the Overkiz side, if the option does not exist...

tetienne (approved):

tetienne:
About the helper, we haven't such thing yet. Do yourself have any otherwise helpers ideas?

iMicknl:
@tetienne actually, in previous PRs I added 2 utils (https://github.com/iMicknl/python-overkiz-api/blob/main/pyoverkiz/utils.py) which are some kind of helpers 😁.

PR #921, opened 2023-08-03
Add StrEnum and Python 3.11 support

iMicknl:


pyproject.toml
```diff
@@ -13,13 +13,14 @@ packages = [
...
 [tool.poetry.dependencies]
-python = ">=3.7,<4.0"
+python = ">=3.8,<4.0"
 aiohttp = "^3.6.1"
 pyhumps = "^3.0.2,!=3.7.3"
 backoff = ">=1.10.0,<3.0"
 attrs = ">=21.2,<24.0"
 boto3 = "^1.18.59"
 warrant-lite = "^1.0.4"
+backports-strenum = "^1.2.4"
```
tetienne, line 23: There is also https://github.com/irgeek/StrEnum

iMicknl: I saw indeed, but seems this is not using the native `StrEnum` for Python 3.11. I rather use the native StrEnum, with a backport for older Python versions.

iMicknl: @tetienne I will add this to a new PR, this PR already got to big with all changes required to 3.11 and CI/CD. :D

tetienne (approved):

PR #923, opened 2023-08-04
Move to StrEnum with backport for older Python versions

iMicknl:


fetzerch:
This PR causes a change of all sensor `unique_id` in https://github.com/home-assistant/core/blob/dev/homeassistant/components/overkiz/entity.py#L117C84-L117C84 between 2023.8.4 (with pyoverkiz 1.9.0) and the current dev (with pyoverkiz 1.10.1). This causes new entity_ids and essentially duplicated and unavailable entities. I'm not sure if that's an intentional breaking change, but I guess it's rather an issue that just slipped in.

The old `Enum` has `__str__` set to [`{classname}.{optionname}`](https://github.com/python/cpython/blob/main/Lib/enum.py#L1187C27-L1187C65) while `StrEnum` uses `{optionvalue}`. E.g `io://XXXX-XXXX-XXXX/XXXXXXX-OverkizState.CORE_DISCRETE_RSSI_LEVEL` became `io://XXXX-XXXX-XXXX/XXXXXXX-core:DiscreteRSSILevelState` (you can double check that in `.storage/core.entity_registry`).

In order to change it back to the old behavior we can either change the integration to use `self.entity_description.key.__class__.__name__.self.entity_description.key.name` in https://github.com/home-assistant/core/blob/dev/homeassistant/components/overkiz/entity.py#L117C84-L117C84 or override `__str__` in the library. I'm not sure if there's any other place that relied on the old behavior but if not it's probably easiest to adapt the integration.

I have not found a corresponding issue in the integration yet. I did find https://github.com/home-assistant/core/issues/99404, but I'm not sure if it's the same.

PR #956, opened 2023-09-02
client: Add refresh_device_states

fetzerch:
`python-overkiz-api` already has a method to refresh all states of all devices. This method is used by the Somfy TaHoma web interface on login/start and it asks all devices to send their states to the gateway.

For some devices like the Somfy Dexxo io garage door motors, this call is essential because these motors do not always send their state updates automatically to the gateway and require a manual refresh. See my discussion with Somfy in https://github.com/Somfy-Developer/Somfy-TaHoma-Developer-Mode/issues/26#issuecomment-1598316333.

Now besides refreshing all states of all devices, the Somfy API also supports to refresh only the states of a specific device via `setup/devices/{device_url}/states/refresh`.

I'll also create a draft PR for `homeassistant/core` that shows how this method could be used to solve #167 (at least for garage doors and gates). But even independent of that, I believe it's a helpful addition to `python-overkiz-api`.

Screenshot of the Somfy API docs:
grafik

### DIFF
--- a/pyoverkiz/client.py
+++ b/pyoverkiz/client.py
@@ -533,6 +533,17 @@ class OverkizClient:
         """
         await self.__post("setup/devices/states/refresh")
 
+    @backoff.on_exception(
+        backoff.expo, NotAuthenticatedException, max_tries=2, on_backoff=relogin
+    )
+    async def refresh_device_states(self, deviceurl: str) -> None:
+        """
+        Ask the box to refresh all states of the given device for protocols supporting that operation
+        """
+        await self.__post(
+            f"setup/devices/{urllib.parse.quote_plus(deviceurl)}/states/refresh"
+        )
+
     @backoff.on_exception(backoff.expo, TooManyConcurrentRequestsException, max_tries=5)
     async def register_event_listener(self) -> str:
         """


fetzerch:
Related PR for homeassistant/core: https://github.com/home-assistant/core/pull/99499

iMicknl:
I will review this + HA PR after my holidays (October). In the past we have already tried to implement such solutions, with a separate task interval in HA. 

The most difficult thing is that users expect this to happen periodically, because they want state updates in HA when they execute a command outside HA. We don't like to poll this endpoint every x minutes, since this is not a best practice according to Somfy.

For TaHoma app it is easy, they just call this endpoint when you open the app. With HA all state updates should happen in the background.

fetzerch:
Hey @iMicknl I know about the problems. It's really a pity that a workaround like this is needed. I replaced my older `Dexxo Pro io` with the newer `Dexxo Smart io` in the hope that this motor wouldn't have the limitation anymore - but it's the same. I was hoping that it's *acceptable* if we (1) limit the polling to the device(s) that really need it and (2) use a relatively high interval like 5m.

Happy to discuss the topic further with you once you're back. I'm slowly migrating over from the [Nymea Somfy integration](https://github.com/nymea/nymea-plugins/tree/master/somfytahoma) that I'm maintaining to Home Assistant and your Somfy integration. Thus the few PRs I've created. Enjoy your vacation! 🌞

Quentame (approved):
LGTM 😌

iMicknl (approved):
Thanks! LGTM

PR #957, opened 2023-09-02
command: Add SET_PARTIAL_POSITION

fetzerch:
The Somfy Dexxo smart io garage door motor supports a third position state: partial. This state can be reached via the 'setPartialPosition' command.

Here's the relevant part from `/setup` for the device:

```json
        {
            "creationTime": 1687080116000,
            "lastUpdateTime": 1687080116000,
            "label": "Garagentor",
            "deviceURL": "io://XXXX-XXXX-XXXX/XXXXXXX",
            "shortcut": false,
            "controllableName": "io:DiscreteGarageOpenerWithPartialPositionIOComponent",
            "definition": {
                "commands": [
                    {
                        "commandName": "addLockLevel",
                        "nparams": 2
                    },
                    {
                        "commandName": "advancedRefresh",
                        "nparams": 2
                    },
                    {
                        "commandName": "close",
                        "nparams": 0
                    },
                    {
                        "commandName": "delayedStopIdentify",
                        "nparams": 1
                    },
                    {
                        "commandName": "down",
                        "nparams": 0
                    },
                    {
                        "commandName": "getName",
                        "nparams": 0
                    },
                    {
                        "commandName": "identify",
                        "nparams": 0
                    },
                    {
                        "commandName": "open",
                        "nparams": 0
                    },
                    {
                        "commandName": "partialPosition",
                        "nparams": 0
                    },
                    {
                        "commandName": "refreshPartialPosition",
                        "nparams": 0
                    },
                    {
                        "commandName": "removeLockLevel",
                        "nparams": 1
                    },
                    {
                        "commandName": "resetLockLevels",
                        "nparams": 0
                    },
                    {
                        "commandName": "setName",
                        "nparams": 1
                    },
                    {
                        "commandName": "startIdentify",
                        "nparams": 0
                    },
                    {
                        "commandName": "stop",
                        "nparams": 0
                    },
                    {
                        "commandName": "stopIdentify",
                        "nparams": 0
                    },
                    {
                        "commandName": "up",
                        "nparams": 0
                    },
                    {
                        "commandName": "wink",
                        "nparams": 1
                    },
                    {
                        "commandName": "runManufacturerSettingsCommand",
                        "nparams": 2
                    },
                    {
                        "commandName": "keepOneWayControllersAndDeleteNode",
                        "nparams": 0
                    },
                    {
                        "commandName": "pairOneWayController",
                        "nparams": 2
                    },
                    {
                        "commandName": "sendIOKey",
                        "nparams": 0
                    },
[...]

PR #972, opened 2023-09-11
command: Fix partial position command

fetzerch:
Commit 4bbc5f8 introduced the `SET_PARTIAL_POSITION="setParitalPosition"` command. However, the command is really called "partialPosition".

---

Sorry, this slipped in - most likely because I was doing the corresponding change in the integration similar to the gate's `SET_PEDESTRIAN_POSITION` and got confused by that. In the pasted JSON in https://github.com/iMicknl/python-overkiz-api/pull/957 you can see that "partialPosition" is correct and not "setPartialPosition".


tetienne (approved):
My bad I also didn't review correctly the previous time.

Thx

PR #1034, opened 2023-11-07
Obfuscate SSID name in diagnostics

iMicknl:


pyoverkiz/obfuscate.py
```diff
@@ -42,7 +42,12 @@ def obfuscate_sensitive_data(data: dict[str, Any]) -> JSON:
         ]:
             data[key] = obfuscate_string(value)
 
-        if value in ["core:NameState", "homekit:SetupCode", "homekit:SetupPayload"]:
+        if value in [
+            "core:NameState",
+            "homekit:SetupCode",
+            "homekit:SetupPayload",
+            "core:SSIDState",
```
tetienne, line 49: What's the value looks like?

tetienne (approved):

pyoverkiz/obfuscate.py
iMicknl: Just the string value of an SSID. But we should filter this out as it can be quite personally. Same as we do for HomeKit codes.

PR #1140, opened 2024-02-14
Add support for ZoneControl 2.0 main probe

Tronix117:


Tronix117:
Thanks @tetienne, however there is an issue with the CI it seems...

PR #1158, opened 2024-02-29
Add enums for Daikin (Altherma 3)

iMicknl:
@jwillemsen fyi. What would be the name of your 'DAIKIN hub'? Is there a specific hub or dongle required.


jwillemsen:
Each Daikin device has a wifi module, some out of the box, some are optional. They all result in the availability of the device in the Daikin Onecta app, maybe call the hub 'daikin_onecta'?

PR #1294, opened 2024-07-12
Add setDateTime command constant

ALERTua:
Adds setDateTime command for Atlantic DHW that sets the datetime for the device.
I have found that my device had the wrong date set on it, so all my experiments with setting away start date and away end date are inconsistent due to this.
I want to make sure that https://github.com/home-assistant/core/pull/121801 sets the correct current datetime on the device before setting away start date and end dates for the user.


ALERTua:
↑ there is no Python 3.12 in the automated checks btw.

iMicknl:
> ↓ there is no Python 3.12 in the automated checks btw.

Thanks @ALERTua! Will need to add this in the coming weeks (and also add this to our package metadata).

PR #1309, opened 2024-07-30
only depend on backports-strenum on Python < 3.11

dotlambda:


iMicknl (approved):
Thanks @dotlambda! Out of interest, did you face any issues without this patch or are you trying to limit the packages you download?

dotlambda:
The latter.

Issue #1323, opened 2024-08-08
Tests are failing with aiohttp 3.10.1


mweinelt:
Hi!

We upgraded aiohttp from 3.9.x to 3.10.1 and started seeing the following test fail on pyoverkiz 1.13.14:

Full log

```
ERROR tests/test_client.py::TestOverkizClient::test_get_api_type_cloud - RuntimeError: no running event loop
ERROR tests/test_client.py::TestOverkizClient::test_get_api_type_local - RuntimeError: no running event loop
ERROR tests/test_client.py::TestOverkizClient::test_get_devices_basic - RuntimeError: no running event loop
ERROR tests/test_client.py::TestOverkizClient::test_fetch_events_basic[events.json-16] - RuntimeError: no running event loop
ERROR tests/test_client.py::TestOverkizClient::test_fetch_events_basic[local_events.json-3] - RuntimeError: no running event loop
ERROR tests/test_client.py::TestOverkizClient::test_fetch_events_simple_cast - RuntimeError: no running event loop
ERROR tests/test_client.py::TestOverkizClient::test_fetch_events_casting[events.json] - RuntimeError: no running event loop
ERROR tests/test_client.py::TestOverkizClient::test_fetch_events_casting[local_events.json] - RuntimeError: no running event loop
ERROR tests/test_client.py::TestOverkizClient::test_get_setup[setup_3_gateways.json-37-3] - RuntimeError: no running event loop
ERROR tests/test_client.py::TestOverkizClient::test_get_setup[setup_cozytouch.json-12-1] - RuntimeError: no running event loop
ERROR tests/test_client.py::TestOverkizClient::test_get_setup[setup_cozytouch_v2.json-5-1] - RuntimeError: no running event loop
ERROR tests/test_client.py::TestOverkizClient::test_get_setup[setup_cozytouch_2.json-15-1] - RuntimeError: no running event loop
ERROR tests/test_client.py::TestOverkizClient::test_get_setup[setup_cozytouch_3.json-15-1] - RuntimeError: no running event loop
ERROR tests/test_client.py::TestOverkizClient::test_get_setup[setup_hi_kumo.json-3-1] - RuntimeError: no running event loop
ERROR tests/test_client.py::TestOverkizClient::test_get_setup[setup_hi_kumo_2.json-3-1] - RuntimeError: no running event loop
[...]

PR #1311, opened 2024-07-30
Drop support for Python 3.8 & 3.9, add support for Python 3.12 in CI/CD and update to latest packages

iMicknl:
This pull request includes multiple changes aimed at updating dependencies, improving type hints, and removing outdated configurations. The most important changes include updating the Python version, enhancing type hints, and cleaning up IntelliJ IDEA project files.

## Breaking

- Renamed private _ssl_context to _ssl: ssl.SSLContext
- Remove support for Python 3.8 and 3.9, bumping the minimum version to Python 3.10

### Dependency and Version Updates:
* [`.devcontainer/devcontainer.json`](diffhunk://#diff-24ad71c8613ddcf6fd23818cb3bb477a1fb6d83af4550b0bad43099813088686L6-L12): Updated the Docker image to `mcr.microsoft.com/devcontainers/python:1-3.12-bullseye` for Python 3.12.
* [`.github/workflows/main.yaml`](diffhunk://#diff-71cabc4177e41ea8f15a89eb65398fa8187739f68a2762706d84885cd8b2ddc3L19-R23): Updated the Python versions in the matrix to include 3.12 and removed 3.8 and 3.9.
* [`.python-version`](diffhunk://#diff-652bad7c1002badcf13f8de3a5096c2724ce54ae36cd68a21ad280547dff8e87L1-R1): Updated the Python version to 3.12.
* [`pyproject.toml`](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L16-R35): Updated dependencies and Python version requirements to `>=3.10,<4.0`.

### Type Hint Improvements:
[...]

PR #1326, opened 2024-08-08
fix tests with aiohttp 3.10

dotlambda:
fixes https://github.com/iMicknl/python-overkiz-api/issues/1323


tetienne (approved):

iMicknl:
Thanks, fixed in #1311

PR #1350, opened 2024-09-01
Update Cozytouch server name to include whitelabeled brands (Thermor, Sauter)

carsso:
Thermor Cozytouch and Sauter Cozytouch are simply whitelabeled Atlantic Cozytouch bridges and apps, using the same servers and accounts.

To ease usage for people buying those whitelabaled products, changing server name to "Cozytouch (Atlantic, Thermor, Sauter)" instead of "Atlantic Cozytouch".
It will be easier to find and choose the right server if you bought a Cozytouch bridge from Thermor or Sauter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ This package is mainly used by Home Assistant Core, to offer the Overkiz integra
 - Hitachi Hi Kumo
 - Nexity Eugénie
 - Rexel Energeasy Connect
+- Sauter Cozytouch
 - Simu (LiveIn2)
 - Somfy Connexoon IO
 - Somfy Connexoon RTS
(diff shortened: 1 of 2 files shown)


iMicknl (commented):
Would it make sense to add extra items to the server list for Thermor_Cozytouch and Sauter_Cozytouch? They are using a different mobile app and this will allow us to add the manufacturer name in as well.

If they switch their servers in the future, we can migrate users without needing any configuration changes.

carsso:
Up to you.
I thought about that possibility before discovering the 3 mobile apps are also using the same backend/user database and my account works exactly the same on all 3 apps.

iMicknl:
My preference would be to have 3 different entries, because (unfortunately) I have seen server migrations / authentication changes in the past.

carsso:
I can update that PR accordingly :) Unless you'd like to do it yourself :) 
No guarantee I'll be able to do it this weekend though.

iMicknl:
Working on a new PR: https://github.com/iMicknl/python-overkiz-api/pull/1358 (since I am now working on a new release for HA anyways ). But your contribution is much valued!

iMicknl:
Replaced by #1358, thanks for your contribution!

PR #1476, opened 2024-12-17
Add actionGroups (scenarios) to diagnostics

iMicknl:


pyoverkiz/client.py
```diff
@@ -460,12 +460,19 @@ async def get_diagnostic_data(self) -> JSON:
...
 
         This data will be masked to not return any confidential or PII data.
         """
-        response = await self.__get("setup")
+        setup = await self.__get("setup")
+        scenarios = await self.__get("actionGroups")
+
+        response = {
+            "setup": obfuscate_sensitive_data(setup),
+            "scenarios": scenarios,  # TODO obfuscate_sensitive_data
```
tetienne, line 472: The `TODO` here is a blocker for me — the docstring above explicitly says *"This data will be masked to not return any confidential or PII data"*, but `actionGroups` payloads carry `deviceURL`, `label`, and `oid` which are exactly what `obfuscate_sensitive_data` exists to scrub.

One wrinkle: `obfuscate_sensitive_data` expects a `dict` at its top level, but the `actionGroups` response is a JSON array. Worth iterating instead:

```suggestion
            "scenarios": [obfuscate_sensitive_data(s) for s in scenarios],
```

Would that work for your use case, or do you need the raw list shape for a specific reason?

pyoverkiz/client.py
```diff
@@ -460,12 +460,19 @@ async def get_diagnostic_data(self) -> JSON:
...
             -> setup devices: <devices>
+            -> scenarios: <actionGroups>
 
         This data will be masked to not return any confidential or PII data.
         """
-        response = await self.__get("setup")
+        setup = await self.__get("setup")
+        scenarios = await self.__get("actionGroups")
+
+        response = {
```
tetienne, line 470: Heads up that this changes the return shape of `get_diagnostic_data()` from a flat dict (keys like `gateways`, `devices`, `rootPlace`) to `{"setup": ..., "scenarios": ...}`. Any downstream caller indexing `data["gateways"]` directly will break silently. Have you had a chance to check the Home Assistant integration side for this?

tetienne (commented):
Nice initiative bringing scenarios into diagnostics — this fills a genuine gap since action groups are a distinct resource not covered by the setup dump. A couple of things to sort out before this is ready though.

PR #1748, opened 2025-09-12
Added support to send multiple commands to multiple devices at once

CZDave:
My change adds support for sending multiple commands to multiple devices at once. I wanted to control multiple windows covers from Home Assistant using Local API and found out that it was slower and was causing issues when I wanted to close all my covers at once (10 devices) for example.

I started working on a modification for the core Overkiz integration in Home Assistant to support this use case from scripts, but realized that I needed to add the support to the API Client first.

Sample usage:
```
client.execute_actions(
    [
        Action(device_url1, [Command(OverkizCommand.CLOSE)]),
        Action(device_url2, [Command(OverkizCommand.CLOSE)]),
    ]
)
```

### DIFF
--- a/pyoverkiz/client.py
+++ b/pyoverkiz/client.py
@@ -679,7 +679,7 @@ class OverkizClient:
         }
         response: dict = await self.__post("exec/apply", payload)
         return cast(str, response["execId"])
-   
+
     @retry_on_auth_error
     async def execute_actions(
         self,
@@ -696,7 +696,7 @@ class OverkizClient:
         }
         response: dict = await self.__post("exec/apply", payload)
         return cast(str, response["execId"])
-      
+
     @retry_on_auth_error
     async def get_scenarios(self) -> list[Scenario]:
         """List the scenarios"""


iMicknl:
Thanks for your contribution @CZDave! Would it be an idea to first implement this in the Home Assistant core integration, and merge this one when you have found a good implementation? This would make it easier to make changes to pyoverkiz if needed, as we prefer to not have breaking changes after release.

You can point Home Assistant to your current feature branch.

iMicknl:
My personal preference is to revamp the action executions in this library anyways, and make it better typed and have a single function that can do one or multiple calls.

iMicknl:
Thanks @CZDave, this would be a good quick workaround, but such additional services won't be accepted in Home Assistant core.

You will need to implement some kind of queuing mechanism in the integration (or underlying package), where all normal HA services/features can be used, but the batching is taken care of automatically.

CZDave:
I see. In that case I think you can close this PR. Thank you for taking your time on this.

I'll see if I can find some time to look into implementing an automatic queueing mechanism. If I'm able to implement that in the integration, would you consider pulling something like this PR to pyoverkiz to support it? Or would you rather have some revamped version as you mentioned first?

iMicknl:
Apologies for the delay @CZDave! I will review this in more detail in the coming weeks, and I am also looking if we can rewrite parts of this library (with breaking changes) for a v2.

There are still a few items where I am not sure yet:
- Should we include a queue in the underlying library, or will this needs to be implemented in the client (e.g. Home Assistant). 
- How should we structure the functions in this library to execute commands? Would this be to difficult (when users have to type Action and Command)

```python3
await execution_apply(
    action=Action(
        device_url="io://0123456789abcdef01234567/12345678",
        commands=[
            Command(name="down"),
            Command(name="refresh")
        ]
    )
)
```

iMicknl:
> Regarding where to add the queue mechanism, in my opinion it should be implemented in the client (HA). Originally I thought about including it in this library in order to make the HA client simpler, but then I thought about the library also being used by other clients that might not be limited in the same way as the HA client, and might want to just send multiple commands directly without a need for the queue at all. So in the end I think the queue usage is actually client specific and should be implemented in the client. But that's just my point of view.

My current thinking is that this would be a great optional feature for the library. It’s not really a Home Assistant limitation, but rather an Overkiz one, since they only allow 10 concurrent executions.

A built-in queue that collects and batches commands could be very useful for many applications. And as long as it’s opt-in, it won’t impact users who don’t need or want this functionality.

---

In the meantime, we’re working on a v2 version of this integration (https://github.com/iMicknl/python-overkiz-api/pull/1872), where we can introduce improvements that may include breaking changes compared to v1.

PR #1866, opened 2025-12-29
Implement ActionQueue for batching actions in OverkizClient

### LINKED ISSUE #1865: Add optional action queue to group executions and avoid Overkiz concurrency limits
Overkiz has strict limits on concurrent executions and the number of queued executions. This makes it difficult for Home Assistant to reliably execute many commands at the same time, especially in scenes or automations that trigger multiple actions simultaneously.

A potential solution is to add an optional queue to the library that groups actions triggered within a short, configurable time window into a single action group. This would reduce the number of concurrent and queued executions sent to Overkiz.

The library already provides execute_action_group(), which supports multiple actions, but users may still execute single actions individually. Additionally, there may be a limit on how many actions can be included in one action group, which should be considered in the design.

## Plan
To mitigate this, we should introduce an optional queue mechanism in the library. This queue would:

- Collect actions that are triggered within a short time window (configurable)
- Automatically group them into a single action group.
- Use a configurable time window, for example under one second, defined when instantiating the client.

iMicknl:
## Enhancement

- Adds ActionQueue for batching device executions (actions)

Fixes #1865

### DIFF
--- a/docs/action-queue.md
+++ b/docs/action-queue.md
@@ -1,6 +1,6 @@
 # Action queue
 
-The action queue automatically groups rapid, consecutive calls to `execute_action_group()` into a single ActionGroup execution. This minimizes the number of API calls and helps prevent rate limiting issues, such as `TooManyRequestsException`, `TooManyConcurrentRequestsException`, or `TooManyExecutionsException`, which can occur if actions are sent individually in quick succession.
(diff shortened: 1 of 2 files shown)


iMicknl:
@copilot please look extensively at all review comments and address them

PR #1872, opened 2025-12-29
2.0.0

### LINKED ISSUE #766: Add default RTS execution duration parameter to Overkiz client
https://github.com/home-assistant/core/blob/ee6886dabde435cd82e69dd6cbefa3dbb15d5122/homeassistant/components/overkiz/executor.py#L80-L89

Migrate Home Assistant code to Overkiz client

### LINKED ISSUE #1183: client.get_setup() - Different responses with Cloud and Local API
Running  client.get_setup() with the cloud API on a Somfy TaHoma server, I get the first item 
`"id": "SETUP-xxx-xxxx-xxxx"`
which seems to be correct.

Running  client.get_setup() with the local API on the Somfy local server, I get the first item 
`"id": "_CountingAttr(counter=82, _default=None, repr=<function obfuscate_id at 0x7fff295e1580>, eq=True, order=True, hash=None, init=True, on_setattr=None, alias=None, metadata={})"`
making the response difficult to exploit as a Json string.

I suspect this behavior not as expected ?
Could you please advise .
Thanks

iMicknl:
# pyoverkiz 2.0.0 Release Notes

PR: https://github.com/iMicknl/python-overkiz-api/pull/1872

## Breaking changes

- Minimum Python version is now Python 3.12 (#1871)
- `client.get_scenarios()` replaced by `client.get_action_groups()` (#1864)
- `Scenario()` model replaced by `ActionGroup()`, where `creation_time` and `metadata` are now optional fields (#1864)
- `client.execute_command()` and `client.execute_commands()` replaced by `client.execute_action_group()` (#1862)
- `OverkizServer` class renamed to `ServerConfig` with additional `server` and `api_type` (cloud/local) property (#1867)
- `generate_local_server` renamed to `create_local_server_config` (#1867)
- `client.api_type` removed and now available via `ServerConfig` (e.g. `client.server_config.api_type`) (#1867)
- The `OverkizClient` constructor now requires passing a `ServerConfig` via `server` (#1867)
- The `OverkizClient` constructor now requires passing a `Credentials` class via `credentials`, e.g. `UsernamePasswordCredentials(USERNAME, PASSWORD)` (#1867)
- The Device class no longer includes the `id` property; `protocol`, `gateway_id`, `device_address`, and `subsystem_id` moved to `device.identifier.{}` (#1930)
- `Device.data_properties` removed (was a non-existent value) (#1930)
- Various `UIClass` and `UIWidget` enums renamed to proper UPPER_SNAKE_CASE (#1925)
- All custom exception classes renamed from `*Exception` to `*Error` (PEP 8); `NotSuchTokenException` fixed to `NoSuchTokenError` (#1987)
- `CommandMode` renamed to `ExecutionMode`, `cancel_command()` to `cancel_execution()`, `execute_scenario()` to `execute_persisted_action_group()`, `execute_scheduled_scenario()` to `schedule_persisted_action_group()` (#1997)
- `Execution.state` is now `ExecutionState` instead of `str`; `ActionGroup.id` and `ActionGroup.oid` are now `str | None`; `get_current_execution` returns `Execution | None` (#2007)
- `get_diagnostic_data()` return shape changed from flat setup dict to `{"setup": ..., "action_groups": ...}` (#2016)
- `get_action_groups()` returns `list[PersistedActionGroup]` instead of `list[ActionGroup]`; `ActionGroup` no longer has `oid`, `creation_time`, or `last_update_time` fields (#2017)
- `ServerConfig.type` renamed to `ServerConfig.api_type` (#1995)
- `Gateway.connectivity` type changed to `Connectivity | None` (was non-optional) (#1994)
- `Gateway.id` and `Place.id` changed from mutable fields to read-only properties (#1994)
- `States.__getitem__` now raises `KeyError` instead of returning `None` — use `States.get()` for old behavior (#1994)
- `CommandDefinitions.__getitem__` now raises `KeyError` instead of returning `None` — use `CommandDefinitions.get()` for old behavior (#1994)
- `Execution.action_group` type changed from `list[Action]` to `ActionGroup | None` (#1994)
- `OverkizClient.check_response` removed and replaced with module-level `pyoverkiz.response_handler.check_response` (#1977)
- `pyhumps` dependency removed and replaced with internal `_case` module (#1988)
- `boto3` and `warrant-lite` moved from required to optional `[nexity]` extra (#2020)
- `event_listener_id` is now a read-only property (#2025)
- `SUPPORTED_SERVERS` is now immutable via `MappingProxyType` (#2025)
- `obfuscate_sensitive_data` returns a new dict instead of mutating (#2025)
- `get_device_definition` returns `Definition` model instead of raw dict (#2025)
- `deviceurl` parameter renamed to `device_url` for consistency (#2025)
- `Location` model uses `None` instead of empty string defaults (#2025)
- Raises `OverkizError` instead of `ValueError` for API data issues (#2025)

## Fixes

- `client.get_current_executions()` now properly typed; `Execution()` model previously returned list type for `action_group` where it should be dict (#1864)
- `Device.data_properties` removed (non-existent value) (#1930)
- `Setup.id` is now `None` for Local API, as it doesn't support this (#1974)
- `get_current_execution` no longer crashes on empty responses (cloud returns `{}`, local returns `[]` or `null`) (#2007)
- Fixed boolean parsing in Cloud API; `"false"` is no longer mapped to `True` (#2015)
- Fixed overly broad `INVALID_FIELD_VALUE` error mapping — now requires "Unable to determine action group setup" substring (14f13ae)
- Removed redundant `_refresh_token_if_expired()` calls in `fetch_events` and `unregister_event_listener` since `_post()` already handles this (14f13ae)
- Fixed `obfuscate_string` regex to properly handle Unicode characters (e.g. accented letters) (#2037)
- Fixed `command_to_enum_name()` to treat digits as word boundaries (e.g. `discover1WayController` → `DISCOVER_1_WAY_CONTROLLER`) (#2032)

## Features

- `client.execute_action_group()` supports multiple execution modes (high priority, internal, geolocated) (#1862)
- `client.execute_action_group()` supports multiple device actions in same request (#1862)
- `OverkizClient` constructor supports passing a `Server` enum directly, e.g. `OverkizClient(server=Server.SOMFY_EUROPE, ...)` (#1867)
- `OverkizClient` now has an optional `ActionQueue` for batching device executions (#1866)
- Device class has helper methods for retrieving state value, commands, and definition (#1930)
- Device class has `identifier` property with protocol, gateway_id, device_address, subsystem_id, base_device_url (#1930)
- Device class includes additional fields (data_properties, ui_profiles, ui_classifiers, attributes) under `device.definition` (#1930)
- `UIClass`, `UIWidget`, and `UIProfile` are auto-generated from server definitions (#1925)
- Enums with `Unknown` value inherit from `UnknownEnumMixin` to reduce repetition (#1925)
- SDK functions to retrieve device info: controllable types, devices, ui_classes, ui_classifiers, ui_profile, ui_widgets (#1925)
- Option to disable PII masking in `get_diagnostic_data` (enabled by default) (#1975)
- `boto3` is now lazy-imported in Nexity Auth Strategy (#1990)
- `start_time`, `execution_type`, `execution_sub_type` fields added to `Execution` model (#2007)
- `NoSuchDeviceError` and `NoSuchActionGroupError` exception types added (#2007)
- `OverkizClientSettings` for RTS default duration and future extensibility (#2018)
- `structure_response` helper consolidates response deserialization (PEP 695 type param syntax) (#2024)
- Firmware management methods (`get_firmware_versions`, `check_firmware_update`, `start_firmware_update`) and `UnsupportedOperationError` exception (ef66c7a)
- Additional sensitive states redacted: `internal:CurrentInfraConfigState`, `core:LocalIPv4AddressState`, `core:IPAddress`, `core:MacAddress`, `core:SerialNumber`, `core:DeviceSerialNumberState` (#2037)
- `lru_cache` added to `camelize_key` for performance parity with `_decamelize_key` (14f13ae)
- `create_local_server_config` now accepts a `server` parameter (#2025)

## Internal / DX

- Migrated from `pyhumps` to internal `_case` module — 8.4x faster decamelize, 94% smaller (#1988)
- Migrated to `cattrs` for centralized model structuring — models are now pure `@define(kw_only=True)` data classes (#1994)
- `boto3` and `warrant-lite` are now optional `[nexity]` dependencies (#2020)
- Added `backoff` `max_time` and `jitter` to all retry decorators (#1993)
- Replaced magic HTTP status codes with `HTTPStatus` constants (b034971)
- Migrated pre-commit to `prek` (Rust-based) (#1887)
- Developer documentation on GitHub Pages: [imicknl.github.io/python-overkiz-api](https://imicknl.github.io/python-overkiz-api/) (#1908)
- Migration guide for upgrading from v1 to v2 (ef2277f)
- Enabled ruff rules: PT (#2000), N (#1987), ERA (#1998 batch), FURB/PERF/RSE/RET/PIE (#2001), PYI (#2003), PTH (#2004), A (#2005), PL (#2025), W, LOG, DTZ, FLY, ISC, PGH, SLF, SLOT, TID, INP, ICN, G, BLE, TRY, Q (#1998)
- Version bumped to `2.0.0rc1` (#2043)


pyoverkiz/auth/strategies.py
```diff
@@ -0,0 +1,471 @@
...
+            username=self.credentials.username,
+            password=self.credentials.password,
+            pool_id=NEXITY_COGNITO_USER_POOL,
+            client_id=NEXITY_COGNITO_CLIENT_ID,
+            client=client,
+        )
+
+        try:
+            tokens = await loop.run_in_executor(None, aws.authenticate_user)
+        except Exception as error:
```
tetienne: **Blocker:** This catches every `Exception` (including `ConnectionError`, `TimeoutError`, `TypeError`, etc.) and converts them all into `NexityBadCredentialsException`. A network timeout or AWS service error would appear as "bad credentials" to the caller, making debugging very difficult.

I would suggest catching only the specific Cognito/warrant auth failure exceptions here and letting other exceptions propagate naturally.

tetienne (commented):
Nice work on this v2 rewrite! The async context manager pattern, strategy/factory auth design, and action queue batching logic are all well-structured. The StrEnum usage is correct for Python 3.12+, and the parametrized fixture tests are thorough.

I went through the codebase with a focus on Python idioms. Some inline comments below, plus a few more that target lines outside the diff:

---

**Blocker: Type annotations don't match runtime behavior**
`pyoverkiz/models.py:43, 89-96` - Multiple fields annotated `str` but default to `None` (e.g. `id: str = field(repr=obfuscate_id, default=None)`, same for `city`, `country`, `postal_code`, `address_line1`, `address_line2`, `longitude`, `latitude`). Downstream code trusting the annotation will skip `None` checks. These should be `str | None`.

**Blocker: `StateDefinition` can crash at runtime**
`pyoverkiz/models.py:379-383` - If both `name` and `qualified_name` are `None`, `self.qualified_name` is never assigned, but the class declares `qualified_name: str` (non-optional) and `get_state_definition` reads it unconditionally. This will raise `AttributeError`. I would suggest either raising `ValueError` in the else branch, or declaring `qualified_name: str | None = None`.

**Suggestion: `field()` in `__init__` parameter defaults**
`pyoverkiz/models.py:58, 111-118, 721` - `field(repr=obfuscate_id, default=None)` used as `__init__` parameter defaults evaluates to an attrs `Attribute` object at definition time, not `None`. If a parameter is ever omitted, the default is the `Attribute` object. What about using plain `None` as parameter defaults and keeping `field()` only on class-level declarations?

**Suggestion: `__getitem__` should raise `KeyError`**
`pyoverkiz/models.py:473-475, 613-615` - `CommandDefinitions.__getitem__` and `States.__getitem__` return `None` on miss. Idiomatic Python expects `KeyError` from `__getitem__`; a separate `get()` method can return `None`. `__contains__` (line 471) relies on the current behavior, so it would need updating too.

**Suggestion: Simplify `bool | None` to `bool`**
`pyoverkiz/client.py:281` - `register_event_listener: bool | None = True`. `None` and `False` behave identically here. `bool = True` would be simpler.

**Question: Is the manual `__init__` pattern on every model intentional?**
Every model uses `@define(init=False)` + hand-written `__init__`, duplicating every field. Was attrs' built-in init with converters or `cattrs` considered? For simple models like `DataProperty`, `Partner`, `Feature`, `ZoneItem`, the manual init adds nothing beyond `self.x = x`.

pyoverkiz/enums/general.py
```diff
@@ -50,7 +39,7 @@ class DataType(IntEnum):
 
 
 @unique
-class FailureType(IntEnum):
+class FailureType(UnknownEnumMixin, IntEnum):
```
tetienne, line 53: Worth noting that `DataType` (lines above) and `ProductType` are the only `IntEnum`s in the enum family that didn't get `UnknownEnumMixin` — you can see the pattern applied here to `FailureType`. An unknown integer from the API (e.g. when Somfy adds a new data type server-side) will raise `ValueError` and crash `fetch_events()` / `get_setup()` for the whole HA integration. Would you consider adding the mixin and an `UNKNOWN = -1` sentinel to `DataType` and `ProductType` as well?

pyoverkiz/client.py
```diff
@@ -799,194 +617,192 @@ async def get_setup_option_parameter(
...
-                raise NoSuchResourceException(message)
+        Returns False if the device does not support this query.
+        """
+        try:
+            response = await self._get(
+                f"setup/devices/{urllib.parse.quote_plus(deviceurl)}/firmwareUpdateCapability"
+            )
+        except UnsupportedOperationError:
+            return False
+        return cast(bool, response["supportsFirmwareUpdate"])
```
tetienne, line 838: These firmware methods are exciting additions, but they're the only new public methods in 2.0 without fixture-based test coverage. This line in particular will raise a `KeyError` if the gateway returns a differently-shaped payload, which won't be very informative for users to debug. Would you be open to adding a fixture and a test before final release, or at least flagging these as experimental in the docstrings for now?

tetienne (commented):
This is a really exciting milestone for the library. The auth strategy factory, `DeviceIdentifier`, `ActionQueue`, `ServerConfig`, and the `*Exception -> *Error` rename all pull in the right direction, and the new docs site is a huge quality-of-life improvement over the README-only approach. The test suite at 311 passing is solid. Really great work getting here.

Before cutting `2.0.0` final, there are a few things I'd want to address:

**Migration guide** - With this many breaking renames (`OverkizServer -> ServerConfig`, `Scenario -> ActionGroup`, `execute_command -> execute_action_group`, `Device.id -> Device.identifier.*`, `CommandMode -> ExecutionMode`, all the `*Exception -> *Error` changes...) the Home Assistant integration maintainers are going to need a single-page "1.x -> 2.0 rename table" to port efficiently. The release drafter generates a PR list, not a cheat sheet. Would a `docs/migration-v2.md` fit here, or is that planned separately?

**DataType / ProductType enums** - These two `IntEnum`s don't have `UnknownEnumMixin`, unlike most of the other enums. An unrecognised integer from the API will raise `ValueError` and crash the HA polling loop. Adding `UnknownEnumMixin` + an `UNKNOWN = -1` sentinel would make them consistent with the rest of the enum family.

**Firmware methods** - The firmware management methods that landed recently have no test coverage yet. Given they're new surface area in a major release, even a small fixture-based test would help. One of them does a bare `response["supportsFirmwareUpdate"]` dict access that'll raise `KeyError` on any differently-shaped response.

**Release process** - Would it make sense to cut a `2.0.0-rc1` first and give the HA Core integration a release cycle to test the migration before tagging final? The breaking change list is substantial enough that an RC seems prudent.

pyoverkiz/enums/general.py
iMicknl: @tetienne I am not sure about this, I don't think they will really change this and otherwise it is a relatively small fix. This constant didn't change in the last 5 years already... Shall we just keep as is?

PR #1893, opened 2026-01-09
refactor(models): use cattrs to reduce boilerplate in model definitions

tetienne:
## Summary

- Adds cattrs as a new dependency to handle structured dict→model conversion
- Simplifies Partner, Connectivity, and Location models by eliminating duplicate field declarations in `__init__` methods
- Reduces ~56 lines of repetitive code while maintaining backward compatibility

## Why

The current models.py has significant boilerplate where fields are declared twice - once in the class body and again in custom `__init__` methods. Using cattrs with `forbid_extra_keys=False` elegantly handles the `**_` pattern for ignoring unknown API fields, allowing attrs to generate the `__init__` automatically.

This is a proof-of-concept that demonstrates the pattern before broader migration to remaining models.

### DIFF
--- a/pyoverkiz/converters.py
+++ b/pyoverkiz/converters.py
@@ -1,0 +1,43 @@
+"""Converters for structuring API data into model instances using cattrs."""
+
+from __future__ import annotations
+
+from enum import Enum, IntEnum
+from typing import TypeVar
+
+import cattrs
+
+from pyoverkiz.enums import StrEnum
+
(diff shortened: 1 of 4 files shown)


iMicknl:
@tetienne I gave it a try: https://github.com/iMicknl/python-overkiz-api/pull/1994/changes. Could you perhaps give it a review? I have also tried a version without cattr, but I do think your suggestion with cattr is indeed better.

PR #1916, opened 2026-01-25
refactor(tests): use explicit assertions following Python idioms

tetienne:
## Summary

Modernizes the test suite with Python best practices:

- Use `is None` / `is not None` for explicit None checks (PEP 8)
- Use `pytest.raises()` pattern for exception tests
- Add ruff `per-file-ignores` for S101 in tests, removing redundant inline `# noqa` comments
- Replace `os.path` with `pathlib.Path` in test_client.py
- Extract shared fixtures (`MockResponse`, `client`, `local_client`) to `tests/conftest.py`

## Test Plan

- [x] All 129 tests pass
- [x] Pre-commit checks pass (ruff, mypy, ty)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,9 @@ select = [
 ]
 ignore = ["E501"] # Line too long
 
+[tool.ruff.lint.per-file-ignores]
+"tests/**" = ["S101"]  # Allow assert in tests
+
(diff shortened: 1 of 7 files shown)


iMicknl:
@tetienne shall we apply these changes on v2/main and close this one? Might be harder to merge this one in, and I am getting closer to V2 release anyways.

PR #1972, opened 2026-04-05
Refactor Nexity authentication to use inline lazy imports

iMicknl:
Refactors the Nexity authentication strategy to load optional dependencies lazily at runtime using standard inline `import` statements inside the `login()` method, rather than importing them at the module level. This keeps the code readable while avoiding unnecessary imports unless Nexity authentication is actually used.

Dependency loading improvements:

* Moved imports of `boto3`, `botocore.config`, `botocore.exceptions`, and `warrant_lite` from the module level into the body of `NexityAuthStrategy.login()`, using a `try/except ImportError` block to raise a clear `NexityServiceException` if the optional dependencies are missing.
* Removed the intermediate `_load_nexity_auth_dependencies` helper function and the `import importlib` module-level import, restoring full type information (no `Any` or `cast` workarounds).

Testing updates:

* Updated tests to patch `boto3.client` and `warrant_lite.WarrantLite` directly, which is the correct approach when the imports live inside the method body rather than at module level.

### DIFF
--- a/pyoverkiz/auth/strategies.py
+++ b/pyoverkiz/auth/strategies.py
@@ -6,17 +6,13 @@ import asyncio
 import base64
 import binascii
 import datetime
+import importlib
 import json
 import ssl
 from collections.abc import Mapping
 from typing import Any, cast
 
-import boto3
 from aiohttp import ClientSession, FormData
-from botocore.client import BaseClient
-from botocore.config import Config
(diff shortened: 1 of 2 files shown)


iMicknl:
Superseeded by https://github.com/iMicknl/python-overkiz-api/pull/1990

PR #1991, opened 2026-04-17
Simplify model definitions with attrs converters and declarative patterns

iMicknl:
## Summary
- Replace manual `__init__` methods with `_flexible_init` decorator + attrs converters across all models (Device, DeviceIdentifier, Location, Setup, Gateway, etc.)
- Add `_to_list`, `_to_optional`, `_to_optional_enum` converter helpers to eliminate per-model boilerplate
- Add dict index to `CommandDefinitions` and `States` containers for O(1) lookups
- Device `ui_class` and `widget` are now public fields resolved at init time (no more private `_ui_class`/`_widget` with property wrappers)
- `Gateway.id` and `Place.id` changed from mutable copies to read-only `@property` aliases
- Remove unused `_LOGGER`, stale pylint comment, redundant one-off converters

Net reduction of ~300 lines while making every model follow the same declarative pattern.

## Breaking changes
- `ServerConfig.type` renamed to `ServerConfig.api_type`
- `Device.ui_class` and `Device.widget` now return `None` instead of raising `ValueError` when no value is available
- `Gateway.id` and `Place.id` are now read-only properties (no longer settable)
- `States.__getitem__` and `CommandDefinitions.__getitem__` now raise `KeyError` instead of returning `None` (use `.get()` for the old behavior)

## Test plan
- [ ] All existing model tests pass
- [ ] Device `ui_class`/`widget` resolve correctly from both direct kwargs and definition fallback
- [ ] States/CommandDefinitions O(1) lookups work correctly
- [ ] Gateway.id and Place.id properties return correct values

### DIFF
--- a/pyoverkiz/models.py
+++ b/pyoverkiz/models.py
@@ -209,13 +209,6 @@ def _to_states(value: list[dict[str, Any]] | States | None) -> States:
     return States(value)
 
 
-def _to_command_definitions(value: Any) -> Any:
-    """Converter: raw list -> CommandDefinitions, or passthrough."""
-    if isinstance(value, list):
-        return _resolve("CommandDefinitions")(value)
-    return value
-
-
 @_flexible_init
 @define(kw_only=True)
 class Device:
--- a/pyoverkiz/models.py
+++ b/pyoverkiz/models.py
@@ -220,26 +220,30 @@ def _to_states(value: list[dict[str, Any]] | States | None) -> States:
     return States(value)
 
 
-def _to_definition(value: dict[str, Any] | Definition) -> Definition:
-    """Converter: raw dict or Definition -> Definition instance."""
-    if isinstance(value, Definition):
-        return value
-    return Definition(**value)
+def _to_command_definitions(value: Any) -> Any:
+    """Converter: raw list -> CommandDefinitions, or passthrough."""
+    if isinstance(value, list):
+        return _resolve("CommandDefinitions")(value)
+    return value
 
 
-@define(init=False, kw_only=True)
+@_flexible_init
+@define(kw_only=True)
 class Device:
     """Representation of a device in the setup including parsed fields and states."""
 
-    attributes: States
+    attributes: States = field(factory=lambda: _to_states(None), converter=_to_states)
     available: bool
     enabled: bool
     label: str = field(repr=obfuscate_string)
     device_url: str = field(repr=obfuscate_id)
     controllable_name: str
-    definition: Definition
-    states: States
-    type: ProductType
+    definition: Definition = field(converter=_to_optional("Definition"))
+    states: States = field(factory=lambda: _to_states(None), converter=_to_states)
+    type: ProductType = field(converter=ProductType)
+    ui_class: UIClass | None = field(default=None, converter=_to_optional_enum(UIClass))
+    widget: UIWidget | None = field(default=None, converter=_to_optional_enum(UIWidget))
+    identifier: DeviceIdentifier = field(init=False, repr=False)
     oid: str | None = field(repr=obfuscate_id, default=None)
     place_oid: str | None = None
     creation_time: int | None = None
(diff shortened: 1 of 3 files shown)


pyoverkiz/models.py
```diff
@@ -850,181 +591,112 @@ def to_payload(self) -> dict[str, object]:
...
+    id: str | None = field(repr=obfuscate_id, default=None)
 
-    def __init__(self, status: str, protocol_version: str, **_: Any):
-        """Initialize Connectivity information."""
-        self.status = status
-        self.protocol_version = protocol_version
+    def __attrs_post_init__(self) -> None:
+        """Resolve id/oid fallback."""
+        if self.oid is None and self.id is None:
+            raise ValueError("Either 'oid' or 'id' must be provided")
```
tetienne, line 619: **Blocker:** This breaks `get_current_executions()` for any active execution. The `actionGroup` embedded in an `Execution` API response has neither `oid` nor `id` — the fixture `tests/fixtures/exec/current-tahoma-switch.json` confirms this. No test exercises this path, which is why CI stays green.

I'd suggest removing the raise and letting both fields be `None` when the action group is embedded in an execution:

```suggestion
        if self.oid is None and self.id is None:
            return
        resolved = cast(str, self.oid or self.id)
        self.id = resolved
        self.oid = resolved
```

If you want to enforce the invariant for standalone action groups (fetched via `get_action_groups()`), that validation would fit better in the client method rather than the model constructor. What do you think?

pyoverkiz/models.py
```diff
@@ -620,44 +221,50 @@ class States:
...
+        """Return the State with the given name or raise KeyError if missing."""
+        return self._index[name]
 
     def __setitem__(self, name: str, state: State) -> None:
         """Set or append a State identified by name."""
-        found = self.__getitem__(name)
-        if found is None:
-            self._states.append(state)
+        if name in self._index:
+            idx = self._states.index(self._index[name])
```
tetienne, line 249: Worth noting: this list scan is O(n) on the update path, which partially defeats the O(1) dict that commit `a8ae00e` introduced for reads. For a state container with tens of items it's fine in practice, but one option is to drop `_states` entirely and iterate the dict values when `__iter__` is needed (Python 3.7+ dicts preserve insertion order). Might simplify the dual-storage invariant as a bonus.

pyoverkiz/models.py
```diff
@@ -30,496 +33,102 @@
...
+    return convert
 
-@define(init=False)
-class CommandDefinitions:
-    """Container for command definitions providing convenient lookup by name."""
 
-    _commands: list[CommandDefinition]
+def _to_str(value: str | None) -> str:
+    """Converter: None -> empty string, otherwise passthrough."""
+    return value or ""
```
tetienne, line 94: The `value or ""` is technically correct for `str | None` inputs (the only falsy str is `""` anyway), but it silently coerces any falsy non-str value (like `0` from a malformed API payload) to `""`. Would this be safer?

```suggestion
    return value if value is not None else ""
```

pyoverkiz/models.py
```diff
@@ -1,9 +1,12 @@
 """Models representing Overkiz API payloads and convenient accessors."""
 
+# mypy: disable-error-code="misc"
```
tetienne, line 3: Have you considered making this narrower? The suppression is file-wide, but it's really only needed for the `@property` definitions on slotted attrs classes (`Gateway.id`, `Place.id`). A targeted `# type: ignore[misc]` on each property would let mypy catch unrelated `misc` errors in the rest of the file.

pyoverkiz/models.py
```diff
@@ -1270,142 +925,99 @@ class ValuePrototype:
...
+class Setup:
+    """Representation of a complete setup returned by the Overkiz API."""
+
+    creation_time: int | None = None
+    last_update_time: int | None = None
+    id: str | None = field(repr=obfuscate_id, default=None)
+    location: Location | None = field(default=None, converter=_to_optional(Location))
+    gateways: list[Gateway] = field(factory=list, converter=_to_list(Gateway))
+    devices: list[Device] = field(factory=list, converter=_to_list(Device))
+    zones: list[Zone] | None = field(default=None, converter=_to_list(Zone))
```
tetienne, line 1019: `_to_list` always returns `[]` on `None` input, so `zones` and `features` will never actually be `None` after construction — but the type annotation says `list[Zone] | None`. Have you considered changing those to `list[Zone]` (no `| None`) to match the actual runtime behavior? Would remove a bunch of dead `if zones is not None:` guards downstream.

(Same applies to `features` on line 1023.)

tetienne (commented):
This is a really impressive refactor — `_flexible_init` cleanly replaces ~40 manual `__init__` methods, the commit history is clean and bisectable, and the -300 LOC net reduction while maintaining the same functionality is a great result. The O(1) dict index in `States` and `CommandDefinitions` is also a meaningful improvement for HA's polling hot path. A few things to look at before merging.

PR #1993, opened 2026-04-17
Improve backoff retry configuration

iMicknl:
## Summary
- Add `__all__` and re-exports to package `__init__.py` for cleaner public API
- Add `max_time` and `jitter` to all backoff retry decorators for more resilient retry behavior

## Test plan
- [ ] Existing tests pass
- [ ] Public API exports are importable from `pyoverkiz`
- [ ] Retry behavior works with new max_time/jitter settings


pyoverkiz/__init__.py
```diff
@@ -3,3 +3,59 @@
...
+    "Command",
+    "Credentials",
+    "Device",
+    "Event",
+    "Execution",
+    "Gateway",
+    "LocalTokenCredentials",
+    "MaintenanceException",
+    "NotAuthenticatedException",
+    "OverkizClient",
```
tetienne: `Server` and `CommandMode`/`ExecutionMode` are missing from the top-level exports. `OverkizClient.__init__` accepts `server: ServerConfig | Server | str`, so users who import from `pyoverkiz` directly will still need to do `from pyoverkiz.enums import Server` while everything else comes from the top level. Have you considered adding `Server` and `ExecutionMode` alongside `ServerConfig`? Or is the intent to keep enums in `pyoverkiz.enums` deliberately?

pyoverkiz/client.py
```diff
@@ -71,11 +71,13 @@ async def refresh_listener(invocation: Details) -> None:
     await _get_client_from_invocation(invocation).register_event_listener()
 
 
-# Reusable backoff decorators to reduce code duplication
+# Reusable backoff decorators with max_tries and max_time to cap total retry duration.
 retry_on_auth_error = backoff.on_exception(
     backoff.expo,
     (NotAuthenticatedException, ServerDisconnectedError),
     max_tries=2,
+    max_time=60,
```
tetienne: Worth noting: with `max_tries=2` there's only a single backoff sleep (~0–1 s with `full_jitter`), so `max_time=60` won't actually cap the retry loop in practice — it'll only fire if a single HTTP request hangs for 60 s. That's still a useful safety net, but it reads as though it bounds the total retry window. A short comment would prevent future confusion?

```suggestion
    max_time=60,  # safety net for a hung request; with max_tries=2 there is only one sleep
```

tetienne (commented):
Really clean PR — the alphabetical `__all__` with re-exports is exactly the kind of public API hygiene that makes a library pleasant to use, and the `full_jitter` addition is a solid thundering-herd fix. A few things worth considering before this lands.

pyoverkiz/client.py
```diff
@@ -84,27 +86,35 @@ async def refresh_listener(invocation: Details) -> None:
...
+    max_time=120,
+    jitter=backoff.full_jitter,
     logger=_LOGGER,
 )
 
 retry_on_too_many_executions = backoff.on_exception(
     backoff.expo,
     TooManyExecutionsException,
     max_tries=10,
+    max_time=300,
```
tetienne, line 112: The execute rate-limit window for the Overkiz gateway is ~28 min 48 s, but `max_time=300` (5 min) will give up well before that window clears. With `max_tries=10` and exponential backoff, retries hit the 300 s ceiling after a few attempts and bail. Worth either aligning `max_time` closer to the rate-limit window, or lowering `max_tries` to avoid the mismatch?

PR #1994, opened 2026-04-17
Migrate to cattrs for centralized model structuring

iMicknl:
## Summary

Alternative to #1991 — replaces per-model converter functions with a centralized `cattrs` Converter.

- Adds `pyoverkiz/converter.py` with hooks for primitive unions, optional attrs classes, enums (with `UnknownEnumMixin`), and custom containers (`States`, `CommandDefinitions`)
- Strips all converter infrastructure from `models.py` — models become pure `@define(kw_only=True)` data classes with only type annotations and defaults
- Updates `client.py` to use `converter.structure()` instead of `Model(**data)` 
- Updates tests to use the converter for API-like dict structuring

### Comparison with #1991

| | #1991 (model-simplifications) | This PR (cattrs) |
|---|---|---|
| **Converter location** | Per-model (`_flexible_init`, `_to_list`, etc.) | Centralized `converter.py` |
| **Model classes** | Still have some converter logic | Pure data classes |
| **Dependencies** | None new | Adds `cattrs` |
| **Forward references** | Eliminated by class reordering | Not needed — cattrs resolves types |
| **Splitting models later** | Requires careful import ordering | No ordering constraints |

## Breaking Changes

- `ServerConfig.type` renamed to `ServerConfig.api_type`.
- `Gateway.connectivity` type changed to `Connectivity | None` (was non-optional).
- `Gateway.id` and `Place.id` changed from mutable fields to read-only properties.
- `States.__getitem__` now raises `KeyError` instead of returning `None` — use `States.get()` for the old behavior.
- `CommandDefinitions.__getitem__` now raises `KeyError` instead of returning `None` — use `CommandDefinitions.get()` for the old behavior.
- `Event.setup_oid` — field name unchanged (cattrs handles the `setupOID` → `setup_oid` conversion automatically).
- `Execution.action_group` type changed from `list[Action]` to `ActionGroup | None`.
- `ActionGroup.label` default simplified (no longer falls back to empty string).

## Test plan
- [x] All 291 tests pass
- [x] ruff check passes
- [x] mypy passes
- [x] ty check passes
- [x] Compare readability and maintainability with #1991


iMicknl:
@copilot resolve the merge conflicts in this pull request

pyoverkiz/converter.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+
+    # Custom container types that take a list in __init__
+    def _structure_states(val: Any, _: type) -> States:
+        if val is None:
+            return States()
+        return States([c.structure(s, State) for s in val])
+
+    def _structure_command_definitions(val: Any, _: type) -> CommandDefinitions:
+        return CommandDefinitions([c.structure(cd, CommandDefinition) for cd in val])
```
tetienne, line 59: **Blocker:** `_structure_states` (the sibling function just above) guards against `None` input, but this one doesn't. If the API returns `"commands": null` for a device, iterating over `None` raises `TypeError`. Worth making these symmetric:

```suggestion
    def _structure_command_definitions(val: Any, _: type) -> CommandDefinitions:
        if val is None:
            return CommandDefinitions()
        return CommandDefinitions([c.structure(cd, CommandDefinition) for cd in val])
```

pyoverkiz/converter.py
```diff
@@ -0,0 +1,64 @@
...
+        return False
+    non_none = [arg for arg in get_args(t) if arg is not type(None)]
+    if any(isinstance(arg, type) and attr.has(arg) for arg in non_none):
+        return False
+    # Exclude pure Optional[Enum] unions — those need the Enum structure hook.
+    return not all(isinstance(arg, type) and issubclass(arg, Enum) for arg in non_none)
+
+
+def _make_converter() -> cattrs.Converter:
+    c = cattrs.Converter()
```
tetienne, line 38: Worth a one-line comment explaining the deliberate choice of `Converter` over `GenConverter`? The key benefit here is that `Converter` silently drops unknown keys (forward-compat with new API fields), while `GenConverter` in strict mode would raise. Future readers might wonder why `GenConverter` wasn't used for the performance benefits.

pyproject.toml
```diff
@@ -19,6 +19,7 @@ dependencies = [
     "attrs>=21.2",
     "boto3<2.0.0,>=1.18.59",
     "warrant-lite<2.0.0,>=1.0.4",
+    "cattrs>=26.1.0",
```
tetienne: Have you checked whether `26.1.0` specifically is required, or whether an older floor would work? Looking at `converter.py`, the hooks used (`register_structure_hook`, `register_structure_hook_func`) have been stable for a long time. A broader floor like `>=23.2` would reduce friction with Home Assistant's dependency tree. What feature from 26.1 are you relying on?

pyoverkiz/client.py
```diff
@@ -655,7 +655,7 @@ async def get_reference_protocol_types(self) -> list[ProtocolType]:
         - label: Human-readable protocol label
         """
         response = await self._get("reference/protocolTypes")
-        return [ProtocolType(**protocol) for protocol in response]
+        return converter.structure(response, list[ProtocolType])
```
tetienne, line 659: Tiny note: this is the only `converter.structure` call in `client.py` that skips `decamelize`. It's intentional (all `ProtocolType` fields are single-word lowercase), but a short comment would prevent a future "fix" PR from adding `decamelize` here for consistency and silently breaking things.

tetienne (commented):
Love the direction here — `converter.py` at 64 lines vs the distributed `_flexible_init` + per-field converter approach is a much cleaner separation of concerns. The models are now pure data classes, which is exactly the right shape. Enum hook correctly routes through `UnknownEnumMixin._missing_`, and the `_is_primitive_union` logic handles the `StateType` passthrough cleanly (and the Optional[Enum] fix in `6701da1` is correct). One small blocker to fix, and a few things worth discussing.

PR #1999, opened 2026-04-19
Enable ruff TC rules (type-checking imports)

iMicknl:
## Summary
- Move type-only imports into `TYPE_CHECKING` blocks across 13 files
- Quote type expressions in `cast()` calls
- Reduces runtime import overhead without changing behavior

## Test plan
- [x] `ruff check .` passes
- [x] `pytest` — 280 tests pass
- [x] `mypy` passes

### DIFF
--- a/pyoverkiz/models.py
+++ b/pyoverkiz/models.py
@@ -4,8 +4,7 @@ from __future__ import annotations
 
 import json
 import re
-from collections.abc import Iterator
-from typing import Any, cast
+from typing import TYPE_CHECKING, Any, cast
 
 from attr import define, field
 
@@ -24,11 +23,15 @@ from pyoverkiz.enums import (
     UpdateBoxStatus,
     UpdateCriticityLevel,
 )
-from pyoverkiz.enums.command import OverkizCommand, OverkizCommandParam
 from pyoverkiz.enums.protocol import Protocol
 from pyoverkiz.enums.server import APIType, Server
 from pyoverkiz.obfuscate import obfuscate_email, obfuscate_id, obfuscate_string
 from pyoverkiz.types import DATA_TYPE_TO_PYTHON, StateType
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from pyoverkiz.enums.command import OverkizCommand, OverkizCommandParam
 
 # ---------------------------------------------------------------------------
 # State & command primitives
@@ -49,7 +52,7 @@ class State:
         if self.type == DataType.NONE:
             return None
         if self.type == DataType.INTEGER:
-            return cast(int, self.value)
+            return cast("int", self.value)
         raise TypeError(f"{self.name} is not an integer")
 
     @property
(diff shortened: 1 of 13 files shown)


iMicknl:
@copilot resolve the merge conflicts in this pull request

pyoverkiz/models.py
```diff
@@ -899,7 +902,10 @@ def __init__(
         **_: Any,
     ) -> None:
         """Initialize ActionGroup from API data and convert nested actions."""
-        self.id = oid or id
+        if oid is None and id is None:
+            raise ValueError("Either 'oid' or 'id' must be provided")
```
tetienne: Caught something here that's a blocker: the `actionGroup` embedded in an `Execution` response (e.g. the in-flight execution fixtures) has neither `oid` nor `id`, so `get_current_executions()` will raise for any active execution. The tests pass because no existing test exercises that path with a real execution payload.

I'd suggest removing this guard (or moving it to the fetch layer for standalone action groups only):

```suggestion
        resolved = self.oid or self.id
        self.id = resolved
        self.oid = resolved
```

What was the original motivation for adding this here — was it a specific API response you ran into?

pyoverkiz/models.py
```diff
@@ -1421,7 +1427,12 @@ def __init__(
 
 @define(kw_only=True)
 class FirmwareStatus:
-    """Firmware status of a device."""
+    """
```
tetienne: The expanded docstring puts the summary on the line after `"""`, which triggers ruff D212. Pre-commit auto-fixes it, but the fix wasn't committed.

Either collapse it to a single line:

```suggestion
    """Firmware status of a device. `up_to_date` indicates whether the current firmware is the latest available version and `notes` contains provider messages about the firmware state."""
```

Or keep it multi-line but move the summary onto the opening line:

```suggestion
    """Firmware status of a device.

    `up_to_date` indicates whether the current firmware is the latest available
    version and `notes` contains provider messages about the firmware state.
    """
```

pyproject.toml
```diff
@@ -132,6 +134,7 @@ ignore = [
     "PLR0913", # Too many arguments — expected for API model constructors
     "PLC0415", # Import not at top level — intentional lazy imports
     "PLR0911", # Too many return statements — acceptable in factory functions
+    "TC006",   # Quoting cast() types hurts IDE autocomplete for no real benefit
```
tetienne, line 138: Noticed a tension here: TC006 is ignored with the rationale "quoting cast() types hurts IDE autocomplete", but the second commit quotes all `cast()` calls anyway (`cast("str", ...)` etc.). These pull in opposite directions. Would you prefer to drop TC006 from the ignore list and enforce quoting consistently, or revert the manual cast quoting and rely on the ignore?

tetienne (commented):
The TYPE_CHECKING moves are all safe — I traced each one and every symbol is used only in annotation contexts, so no runtime impact. And the `full_jitter` + `max_time` additions in the earlier commits are a nice improvement. Two things caught my eye that are worth fixing before this merges though.

iMicknl:
@tetienne this conflicts quite a bit with the `cattr` changes, and the benefit is not too much. I will close this.

PR #2015, opened 2026-04-24
Refactor boolean parsing to use a dedicated function and add tests

iMicknl:


iMicknl:
@tetienne quite interesting that we never spotted this bug earlier :D

PR #2016, opened 2026-04-24
Add action groups to diagnostics

iMicknl:
## Summary

- `get_diagnostic_data()` now fetches both `/setup` and `/actionGroups` endpoints
- Returns a structured dict with named sections instead of a flat setup dump
- `obfuscate_sensitive_data()` extended to accept lists, so action groups are properly obfuscated

## Breaking changes

- `get_diagnostic_data()` return shape changed from flat setup dict to `{"setup": ..., "action_groups": ...}`
- Callers accessing `data["gateways"]` must update to `data["setup"]["gateways"]`
- Return type narrowed from `JSON` to `dict[str, Any]`

## Test plan

- [x] New test for list obfuscation in `test_obfuscate.py`
- [x] New test for structured return shape in `test_client.py`
- [x] 3 existing diagnostic tests updated for new shape
- [x] Full suite passes (329/329)


iMicknl:
@copilot fix prek (linting)

